PR #28, opened 2015-02-09
Calls to sntp adjusted to syntax from ntp 4.2.8 (bnc#916617)

ancorgs:
Merging suggestions from https://bugzilla.suse.com/show_bug.cgi?id=916617 and the code already existing in SLE12 (for ntp 4.2.6).


jreidinger:
this is a backport? I think I see similar change in past.
otherwise looks good

ancorgs:
Kind of a backport. For SLE12/openSUSE we adjusted to code to ntp 4.2.6, here
https://github.com/yast/yast-ntp-client/blob/master/src/clients/ntp-client_proposal.rb#L405

And this is adjusting it to 4.2.8 (looks like the ntp developers enjoy changing the syntax in every version).

jreidinger:
ok, lgtm

ancorgs:
I can merge, but I would need a YCP veteran to submit it :wink:

PR #30, opened 2015-06-26
Fix missing ntp

jreidinger:


imobach:
LGTM

src/clients/ntp-client_auto.rb
```diff
@@ -76,9 +76,15 @@ def main
         #               cloning the just installed system
         if Mode.config && Stage.initial
           ntp_conf = "/etc/ntp.conf"
+          installed_ntp_conf = File.join(Installation.destdir, ntp_conf)
+          # if ntp is not installed at all we cannot copy it, so return empty values
+          if !::FileUtils.exist?(installed_ntp_conf)
+            Builtins.y2milestone("Ntp is not installed, so return empty hash")
```
kobliha: Builtins :)?

jreidinger: well, my goal is to make small fix. If I plan to use logger, then I probably ends up using auto client, which will be more work and deserve scrum item :)...or do I overlook typo?

kobliha: No typo I'd know of. But just that we should omit using B*s. It's not a hard requirement though.

PR #31, opened 2015-07-13
chronos.cru.fr is not realy free accessable: Access policy : accès sous conditions

varkoly:
chronos.cru.fr is not realy free accessable: Access policy : accès sous conditions


mvidner:
Thanks, merged with an adjusted version + changelog.

PR #32, opened 2015-07-20
Calls to sntp adjusted to the syntax of ntp 4.2.8 (bnc#937837)

ancorgs:


src/clients/ntp-client_proposal.rb
```diff
@@ -399,12 +399,13 @@ def Write(param)
 
         Builtins.y2milestone("Running sntp to sync with %1", ntp_server)
 
-        # -s: do set the system time
+        # -S: do set the system time
```
jreidinger: np: why not simple `set the system time`

src/clients/ntp-client_proposal.rb
```diff
@@ -399,12 +399,13 @@ def Write(param)
...
         # -t 5: timeout of 5 seconds
         # -l <file>: log to a file to not mess text mode installation
+        # -c: causes all IP addresses to which ntp_server resolves to be queried in parallel
```
jreidinger: np: maybe simple `querry all ip addresses in parallel`

jreidinger:
LGTM

ancorgs:
I took the descriptions for the comments directly from the bug reporter, who turns out to be the ntp maintainer. I think they are good enough.

PR #33, opened 2015-08-04
Moved a file from this package to yast2-country.

mvidner:
Companion to https://github.com/yast/yast-country/pull/55
Moved
 from yast2-ntp-client:clients/ntp-client_proposal.rb
 to   yast2-country:   clients/timezone_ntp.rb
                       lib/country/ntp_adapter.rb

because the file provides a part(!) of a dialog for the time settings,
and the interface has grown too tangled it does not make sense to
keep them in 2 packages/repos.

Yes, now the problem has been shifted a level lower,
to the NtpClient interface

Also, despite being named *_proposal, the client has never been
used in the proposal framework.


mvidner:
I intend to make a better refactoring than this one after all.

PR #34, opened 2015-08-06
Ensure ntp is installed (when Install Recommended Packages is off)

mvidner:
This is a much shorter version ;-) of https://github.com/yast/yast-country/pull/55 and #33.

https://bugzilla.suse.com/show_bug.cgi?id=938951

### DIFF
--- a/src/clients/ntp-client_proposal.rb
+++ b/src/clients/ntp-client_proposal.rb
@@ -369,17 +369,11 @@ module Yast
       required_package = "ntp"
 
       #In 1st stage, schedule packages for installation
-      #but not in case user wants to set the time only (F#302917)
-      #(ntpdate is in inst-sys so we don't need the package)
-      if Stage.initial && !ntpdate_only
+      if Stage.initial
         Yast.import "Packages"
         Packages.addAdditionalPackage(required_package)
-        # bugzilla #327050
-        # Agent for writing /etc/ntp.conf needs to be installed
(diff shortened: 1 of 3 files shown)


src/clients/ntp-client_proposal.rb
```diff
@@ -369,17 +369,11 @@ def Write(param)
       required_package = "ntp"
```
ancorgs: Better as constant?

ancorgs:
Apart from a NP comment, looks good to me.

kobliha:
@mvidner NOTLGTM :(

/usr/share/YaST2/modules/Packages.rb.addAdditionalPackage says:

"OBSOLETE! Please, use PackagesProposal::AddResolvables() instead"

kobliha:
There is a reason for that: when users go back, they can decide not to use ntp and then you have to call: PackagesProposal::RemoveResolvables()

mvidner:
@kobliha I can easily replace `Packages.addAdditionalPackage` with `PackagesProposal.AddResolvables`. But without the accompanying `PackagesProposal.RemoveResolvables` it will not be helpful. And the surrounding code has no easy way to accommodate RemoveResolvables, so we would be only papering over the problem.

mvidner:
Thanks for pointing out the obsolete API. I've tried to replace it but it seems that the current PR is better off as is.

PR #35, opened 2015-08-13
Offer pool.ntp.org by default

ancorgs:


src/NtpClient.ycp
```diff
@@ -281,6 +281,8 @@ global list GetNtpServersByCountry ( string country, boolean terse_output ) {
         country_names = GetCountryNames ();
     }
 
+    boolean default_already_chosen = false;
+    boolean selected = false;
```
lslezak: NP: You do not need to keep the `selected` value between the iterations or after the `maplist` call, it would be better to move it inside the iteration to have a smaller scope.

lslezak:
LGTM

PR #36, opened 2015-08-17
Offer pool.ntp.org by default (bnc#940881)

ancorgs:
- This is a direct port to ruby of https://github.com/yast/yast-ntp-client/pull/35 (fix already verified by reporter)
- This PR also includes two commits made by @schubi2 to the SLES-12-GA branch (notice the difference with SLE-12-GA, which is the correct one). I'll delete the branch with the extra S after merging.


imobach:
LGTM

PR #41, opened 2015-11-13
L3: Also remove the "restrict" server option (bsc#954442)

lslezak:
...from `/etc/ntp.conf` when a server is removed, do not leave unused values.

- 2.17.19

### The Original Bug

After adding an NTP server in the YaST module the `/etc/ntp.conf` file contains both `server` and `restrict` options:

    # yast2 ntp-client
    # grep ntp.suse.cz /etc/ntp.conf
    server ntp.suse.cz iburst
    restrict ntp.suse.cz  

After removing the NTP server in the YaST module the `restrict` option is left:

    # yast2 ntp-client 
    # grep ntp.suse.cz /etc/ntp.conf
    restrict ntp.suse.cz

### Tested Fix

Add the server again:

    # yast2 ntp-client
    # grep ntp.suse.cz /etc/ntp.conf
    server ntp.suse.cz iburst
    restrict ntp.suse.cz  

After removing the NTP server in the YaST module now the `restrict` option is gone as well:

    # yast2 ntp-client 
    # grep ntp.suse.cz /etc/ntp.conf
    #


jreidinger:
LGTM

lslezak:
Thanks, I'll merge it after I get the fix confirmation from the customer.

src/NtpClient.ycp
```diff
@@ -1084,14 +1084,50 @@ global define integer findSyncRecord (string type, string address) ``{
 }
 
 /**
+ * Remove the "restrict" option for the server if it is not used anymore (bsc#954442)
+ * @param string removed_server Address of the server
+ */
+define void removeRestrictServerOption(string removed_server)
+{
+    if (removed_server != "")
```
jreidinger: what about passed nil? Also early exit will look better like
```
if (removed_server == nil || removed_server == "")
  return;
...
```

src/NtpClient.ycp
```diff
@@ -1107,8 +1143,23 @@ global define boolean deleteSyncRecord (integer index) ``{
...
     }
+
+    y2milestone("Removing record at index %1: %2", index, ntp_records[index]:nil);
+
+    // server removal?
+    if (ntp_records[index,"type"]:"" == "server")
+    {
+        string address = ntp_records[index,"address"]:"";
+
+        if (address != "")
```
jreidinger: RemoveRestrictServerOption can handle `""` so duplicite check here

jreidinger:
LGTM

PR #46, opened 2016-01-22
Bnc 960455 sle11

schubi2:
Sntp uses '-K /dev/null' because the kod file could not exist.


src/ntp-client_proposal.ycp
```diff
@@ -238,11 +238,12 @@ without having package %1 installed"), required_package));
 
 	// -S: do set the system time
 	// -t 5: timeout of 5 seconds
+        // -K /dev/null: use /dev/null as KoD history file (if not specified, /var/db/ntp-kod will be used and it doesnt exist)
```
imobach: doesnt --> doesn't

imobach:
Just a minor typo. Otherwise, LGTM.

PR #47, opened 2016-04-14
Calls to sntp adjusted to syntax from ntp 4.2.8 (FATE#320394)

teclator:
The original fix was done here #28 and here #46. So, it could be considered a backport.


imobach:
LGTM

PR #48, opened 2016-04-14
Calls to sntp adjusted to syntax from ntp 4.2.8 (FATE#320394)

teclator:
It is a backport from #28 and #46.


VERSION
```diff
@@ -1 +1 @@
-2.17.14
+2.17.14.1
```
imobach: Same version than Code-11-SP1? Check #47.

imobach: Ok, so no difference between both branches.

imobach:
LGTM

VERSION
teclator: No, it is the same code for both and diverge in SP3.

PR #50, opened 2016-04-15
Calls to sntp adjusted to the syntax of ntp 4.2.8 (FATE#320392)

teclator:
The fix was original added here #32, and also added a fix from here #46.


imobach:
LGTM

PR #53, opened 2016-05-06
Fixed the default synchronization interval in the help text (bsc#978453)

lslezak:
Originally I wanted to simply fix a trivial [bug 978453](https://bugzilla.suse.com/show_bug.cgi?id=978453) - the default value in the help text did not match the value in the UI, but after looking into it I found several hidden skeletons or zombies...

# Changes

- Unify the default value in the code (BTW the proposal client used   a different default)
- Set the initial state of the interval widget according to the current synchronization mode
- Fixed firewall widget layout (double frame widget)
- 3.1.23

# Screenshots

Here are the screenshots showing the fixes in the UI.

### Original State

After starting the module the *Inverval of the Synchronization in Minutes* widget was enabled although the initial start mode was *Now and on Boot*. The setting makes sense only in the *Synchronize without Daemon* mode.

If you change the mode to something else and then set it back then the widget is properly disabled. So changing the mode is handled properly, only the initial state was not set correctly:

ntp_orig_init

### Fixed

After fix the initial state is correctly set to disabled when the start mode is *Now and on Boot*:

ntp_fix_init

### Original State

Originally there were displayed two `Frame` widgets around the Firewall settings:

ntp_orig_sec

### Fixed

The fix removed one of them:

ntp_fix_sec

### DIFF
--- a/src/include/ntp-client/helps.rb
+++ b/src/include/ntp-client/helps.rb
@@ -8,6 +8,7 @@
 # $Id$
 module Yast
   module NtpClientHelpsInclude
+    Yast.import "NtpClient"
     def initialize_ntp_client_helps(include_target)
       textdomain "ntp-client"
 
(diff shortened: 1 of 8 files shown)


src/include/ntp-client/helps.rb
```diff
@@ -31,16 +32,14 @@ def initialize_ntp_client_helps(include_target)
...
         "start"              => _(
           "<p><b><big>Start NTP Daemon</big></b><br>\n" +
             "Select whether to start the NTP daemon now and on every system boot. \n" +
-            "The NTP daemon resolves host names when initializing. Your\n" +
-            "network connection must be started before the NTP daemon starts.</p>\n" +
-	    "Selecting <b>Synchronize without Daemon</b> the ntp daemon will not be activated. \n" +
-	    "The system time will be set periodically. The interval is configurable. It is 15 minutes by default.\n " +
-	    "You can change this when the system was set up."
-	),
+	    "Selecting <b>Synchronize without Daemon</b> the ntp daemon will not be activated\n" +
```
imobach: np: I would upcase NTP

imobach:
Just a minor note. Otherwise, LGTM.

PR #54, opened 2016-05-19
Backport from SP1

schubi2:


src/clients/ntp-client_auto.rb
```diff
@@ -76,9 +76,15 @@ def main
         #               cloning the just installed system
         if Mode.config && Stage.initial
           ntp_conf = "/etc/ntp.conf"
+          installed_ntp_conf = File.join(Installation.destdir, ntp_conf)
+          # if ntp is not installed at all we cannot copy it, so return empty values
+          if !::FileUtils.exist?(installed_ntp_conf)
```
mvidner: Uh, something is not right:
```console
$ ruby -rfileutils -e 'puts ::FileUtils.exist? "/etc"'
-e:1:in `<main>': undefined method `exist?' for FileUtils:Module (NoMethodError)
```

imobach:
LGTM. The Jenkins CI is failing because `ci.opensuse.org` cannot reach `api.suse.de`, but Travis is running fine.

src/clients/ntp-client_auto.rb
mvidner: For reference, it it is a mistake in the original fix #30, later fixed in https://github.com/yast/yast-ntp-client/commit/f6e154482df4b1db66022cfe821c5a7f43e867d2.

imobach:
Thanks @mvidner for noticing the problem. LGTM.

PR #55, opened 2016-05-23
Calls to sntp adjusted to syntax from ntp 4.2.8 (bsc#979421)

teclator:
I have updated the syntax of sntp call, but also we check now if the server is not reachable checking the presence of "lookup error" in the stderr **[[1]]**.

screenshot_opensuse13_2016-05-23_07 36 48

screenshot_opensuse13_2016-05-23_07 36 28

I added this check because sntp always returns 0 although the server is not answering or not reachable, so the current NTP Test is useless.

We could do more checks like for example check if stdout includes "no UCST reponse" **[[2]]** and maybe other checks, but i'm afraid the answer could change or be very dependent of current responses what could need updates each time the response changes but on the other hand this is only a Test functionality and for this reason not a critical one.

[1]: https://github.com/ntp-project/ntp/blob/1a399a03e674da08cfce2cdb847bfb65d65df237/sntp/main.c#L461
[2]: https://github.com/ntp-project/ntp/blob/1a399a03e674da08cfce2cdb847bfb65d65df237/sntp/main.c#L791

### DIFF
--- a/package/yast2-ntp-client.changes
+++ b/package/yast2-ntp-client.changes
@@ -2,7 +2,7 @@
 Mon May 23 06:19:31 UTC 2016 - kanderssen@suse.com
 
 - Calls to sntp adjusted to the syntax of ntp 4.2.8
-  (bsc#bsc#979421)
+  (bsc#979421)
 - 3.1.12.1
 
 -------------------------------------------------------------------
--- a/package/yast2-ntp-client.changes
+++ b/package/yast2-ntp-client.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon May 23 06:19:31 UTC 2016 - kanderssen@suse.com
+
+- Calls to sntp adjusted to the syntax of ntp 4.2.8
+  (bsc#bsc#979421)
+- 3.1.12.1
+
+-------------------------------------------------------------------
 Tue Sep  2 10:00:18 UTC 2014 - cwh@suse.com
 
 - Using Report.Error to be autoyast-proof
--- a/src/modules/NtpClient.rb
+++ b/src/modules/NtpClient.rb
@@ -1046,36 +1046,34 @@ module Yast
     end
 
 
-    # Test if specified NTP server answers
+    # Test if a specified NTP server is reachable by IPv4 or IPv6 (bsc#74076),
+    # Firewall could have been blocked IPv6
     # @param [String] server string host name or IP address of the NTP server
     # @return [Boolean] true if NTP server answers properly
     def reachable_ntp_server?(server)
-      # testing the server using IPv4 and then using IPv6 protocol
-      # bug #74076, Firewall could have been blocked IPv6
-      return true if sntp_test(server) == 0
+      return true if sntp_test(server)
 
-      sntp_test(server, 6) == 0
+      sntp_test(server, 6)
     end
 
-    # -K /dev/null: use /dev/null as KoD history file (if not specified,
-    #               /var/db/ntp-kod will be used and it doesn't exist)
-    # -c: concurrently query all IPs; -t 5: five seconds of timeout
+    # Test NTP server answer for a given IP version.
     # @param [String] server string host name or IP address of the NTP server
     # @param [Fixnum] integer ip version to use (4 or 6)
     # @return [Fixnum] sntp exit code or 1 in case of lookup error
     def sntp_test(server, ip_version = 4)
       output = SCR.Execute(
         path(".target.bash_output"),
+        # -K /dev/null: use /dev/null as KoD history file (if not specified,
+        #               /var/db/ntp-kod will be used and it doesn't exist)
+        # -c: concurrently query all IPs; -t 5: five seconds of timeout
         "LANG=C /usr/sbin/sntp -#{ip_version} -K /dev/null -t 5 -c #{server}"
       )
 
       Builtins.y2milestone("sntp test response: #{output}")
       # sntp returns always 0 if not called with option -S or -s (set system time)
-      # so this is a workaround at least to return 1 in case server is not
+      # so this is a workaround at least to return false in case server is not
       # reachable.
-      return 1 if output["stderr"].include?("lookup error")
-
-      output["exit"]
+      output["stderr"].include?("lookup error") ? false : output["exit"] == 0
     end
 
     # Handle UI of NTP server test answers
--- a/src/modules/NtpClient.rb
+++ b/src/modules/NtpClient.rb
@@ -1070,9 +1070,13 @@ module Yast
       )
 
       Builtins.y2milestone("sntp test response: #{output}")
+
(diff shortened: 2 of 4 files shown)


src/modules/NtpClient.rb
```diff
@@ -1045,42 +1045,58 @@ def Summary
...
+      output = SCR.Execute(
+        path(".target.bash_output"),
+        "/usr/sbin/sntp -#{ip_version} -K /dev/null -t 5 -c #{server}"
+      )
+
+      Builtins.y2milestone("sntp test response: #{output}")
+      # sntp returns always 0 if not called with option -S or -s (set system time)
+      # so this is a workaround at least to return 1 in case server is not
+      # reachable
+      return 1 if output["stderr"].include?("lookup error")
```
ancorgs: To check the content of the output you should first ensure it will be in English appending `LANG=C` to the command line call.

src/modules/NtpClient.rb
```diff
@@ -1045,42 +1045,58 @@ def Summary
...
-      if verbosity != :no_ui
-        UI.OpenDialog(
-          # An informative popup label diring the NTP server testings
-          Left(Label(_("Testing the NTP server...")))
-        )
-      end
+      return reachable_ntp_server?(server) if verbosity == :no_ui
+
+      UI.OpenDialog(
+        # An informative popup label diring the NTP server testings
```
ancorgs: diring => during.

On the other hand, we maybe have something more specific to report progress that `OpenDialog`. Let me check the `Report` module...

teclator: Nice. This PR is for 13.2, for master i will add also some of the questions asked here about take care of stdout and stderr and we could try to improve the interface.

src/modules/NtpClient.rb
```diff
@@ -1045,42 +1045,58 @@ def Summary
...
     def TestNtpServer(server, verbosity)
-      if verbosity != :no_ui
-        UI.OpenDialog(
-          # An informative popup label diring the NTP server testings
-          Left(Label(_("Testing the NTP server...")))
-        )
-      end
+      return reachable_ntp_server?(server) if verbosity == :no_ui
+
+      UI.OpenDialog(
```
ancorgs: I found it. The right way of doing it would be `Popup.Feedback`. Something like (not tested):
```ruby
ok = false
Popup.Feedback(_("Testing the NTP server..."), Message.takes_a_while) do
  Builtins.y2milestone("Testing reachability of server %1", server)
  ok = reachable_ntp_server?(server)
end
```

ancorgs: Not saying your version is wrong, just that I like more using `Feedback`

package/yast2-ntp-client.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon May 23 06:19:31 UTC 2016 - kanderssen@suse.com
+
+- Calls to sntp adjusted to the syntax of ntp 4.2.8
+  (bsc#bsc#979421)
```
ancorgs: bsc#bsc#bsc :-)

src/modules/NtpClient.rb
```diff
@@ -1045,42 +1045,53 @@ def Summary
...
+      sntp_test(server, 6) == 0
+    end
+
+    # -K /dev/null: use /dev/null as KoD history file (if not specified,
+    #               /var/db/ntp-kod will be used and it doesn't exist)
+    # -c: concurrently query all IPs; -t 5: five seconds of timeout
+    # @param [String] server string host name or IP address of the NTP server
+    # @param [Fixnum] integer ip version to use (4 or 6)
+    # @return [Fixnum] sntp exit code or 1 in case of lookup error
+    def sntp_test(server, ip_version = 4)
```
ancorgs: Thinking twice about this method. It's strange to have a method that returns a possibly modified exit code. At the end of the day, we are always comparing it's output with 0, so I would make it boolean.

A method that returns "low level stuff" like exit codes should not mess with that info. It's unexpected.

src/modules/NtpClient.rb
```diff
@@ -1045,42 +1045,53 @@ def Summary
...
+    # @return [Boolean] true if NTP server answers properly
+    def reachable_ntp_server?(server)
+      # testing the server using IPv4 and then using IPv6 protocol
+      # bug #74076, Firewall could have been blocked IPv6
+      return true if sntp_test(server) == 0
+
+      sntp_test(server, 6) == 0
+    end
+
+    # -K /dev/null: use /dev/null as KoD history file (if not specified,
```
ancorgs: These comments about sntp syntax are probably better placed before the call, rather than in the method description. They are implementation details of the test.

src/modules/NtpClient.rb
```diff
@@ -1046,36 +1046,34 @@ def Summary
...
 
       Builtins.y2milestone("sntp test response: #{output}")
       # sntp returns always 0 if not called with option -S or -s (set system time)
-      # so this is a workaround at least to return 1 in case server is not
+      # so this is a workaround at least to return false in case server is not
       # reachable.
-      return 1 if output["stderr"].include?("lookup error")
-
-      output["exit"]
+      output["stderr"].include?("lookup error") ? false : outpu["exit"] == 0
```
ancorgs: `outpu["exit"]` missing `t`

ancorgs:
I miss a comment saying that we left other checks out for robustness reasons. Mentioning the no UCST reponse" explicitly, as you did in the pull request description.

src/modules/NtpClient.rb
```diff
@@ -1045,42 +1045,55 @@ def Summary
...
+    def reachable_ntp_server?(server)
+      return true if sntp_test(server)
+
+      sntp_test(server, 6)
+    end
+
+    # Test NTP server answer for a given IP version.
+    # @param [String] server string host name or IP address of the NTP server
+    # @param [Fixnum] integer ip version to use (4 or 6)
+    # @return [Fixnum] sntp exit code or 1 in case of lookup error
```
ancorgs: Outdated doc

src/modules/NtpClient.rb
```diff
@@ -1045,42 +1045,55 @@ def Summary
...
-    # Test if specified NTP server answers
+
+    # Test if a specified NTP server is reachable by IPv4 or IPv6 (bsc#74076),
+    # Firewall could have been blocked IPv6
+    # @param [String] server string host name or IP address of the NTP server
+    # @return [Boolean] true if NTP server answers properly
+    def reachable_ntp_server?(server)
+      return true if sntp_test(server)
+
+      sntp_test(server, 6)
```
ancorgs: Now (after redefining `sntp_test` and relocating the comments about the firewall) it looks more reasonable to reduce the body of this method to.
```ruby
sntp_test(server) || sntp_test(server, 6)
```

ancorgs:
LGTM. Sorry for the slow stream of comments... but you fix stuff too fast :-)

PR #56, opened 2016-05-23
Calls to sntp adjusted to the syntax of ntp 4.2.8 (bsc#979416)

teclator:
Replaced ntpdate command by sntp and take in account if stderr contains "lookup error", see #55 to know more.


ancorgs:
LGTM

PR #57, opened 2016-05-24
Check if server is not reachable (bsc#977610)

teclator:
It started from this PBI [trello card](https://trello.com/c/jUAYDPQD/891-sles12-sp2-p3-977610-yast2-ntp-client-configuration-returns-always-successful-message-for-test-public-ntp-server) and as consequence of [(bsc#977610)](https://bugzilla.suse.com/show_bug.cgi?id=977610), but some code is very old and not well tested.

Taking this in account old testsuite tests have been replaced by Rspec ones and new methods are already tested, i have also enable rubocop and test coveralls.


package/yast2-ntp-client.changes
```diff
@@ -1,4 +1,14 @@
 -------------------------------------------------------------------
+Thu Jun  2 07:40:56 UTC 2016 - kanderssen@suse.com
+
+- Fixed Ntp Test adding a workaround to return false in case given
+  server is not reachable (bsc#977610)
+- Replaced old testsuite tests with Rspec tests
```
imobach: "old testsuite with RSpec tests"

package/yast2-ntp-client.spec
```diff
@@ -47,11 +49,14 @@ This package contains the YaST2 component for NTP client configuration.
 %prep
 %setup -n %{name}-%{version}
 
+%check
+rake test:unit
+
 %build
-%yast_build
+yardoc
```
imobach: Please, remove this line as we're not building _devel-doc_ packages anymore.

.rubocop.yml
```diff
@@ -0,0 +1,122 @@
...
+
+Style/FileName:
+  Include:
+    - 'src/lib/**/*.rb' # force correct names on new code
+
+Style/HashSyntax:
+  Include:
+    - 'src/lib/**/*.rb' # force new hash style only for new code
+
+# JR: I found lambda easier to read and understand then ->()
```
jreidinger: ha, I see my opinions lives :wink:

src/include/ntp-client/dialogs.rb
```diff
@@ -124,12 +123,12 @@ def SimpleDialog
...
 
       startInit(nil)
-      CWM.handleWidgets(w, { "ID" => "never" })
+      CWM.handleWidgets(w, "ID" => "never")
 
       ret = CWM.Run(
         w,
         # yes-no popup
-        { :abort => fun_ref(method(:reallyExitSimple), "boolean ()") }
+        :abort => fun_ref(method(:reallyExitSimple), "boolean ()")
```
jreidinger: rubocop do not complain with using old hash syntax?

src/modules/NtpClient.rb
```diff
@@ -1263,6 +1213,44 @@ def AutoPackages
...
 
+  private
+
+    def read_known_servers
+      servers_file = Directory.find_data_file("ntp_servers.yml")
+
+      return {} if !servers_file
+
+      # FIXME: This is wrong approach, because datadir should be array of all Y2DIR/DATA
+      # so we need to skip it in tests, even if we set Y2DIR properly
```
jreidinger: is it true? I think that find_data_file handle Y2DIT, not?

src/modules/NtpClient.rb
```diff
@@ -1263,6 +1213,44 @@ def AutoPackages
...
+      servers = YAML.load_file(servers_file)
+      if servers.nil?
+        log.error("Failed to read the list of NTP servers")
+        return {}
+      end
+
+      log.info "#{servers.size} known NTP servers read"
```
jreidinger: I prefer to have in log real vaues instead of number, it is not much helpful for debugging

test/REPORT
```diff
@@ -0,0 +1,3 @@
+I finally finished this PBI, thanks Ladislav for a nice UI review and suggestions, also Jozef for check if some openQA adapt was needed.
+
+I'm working now in this PBI adding rspec tests to it.
```
jreidinger: ???

.rubocop.yml
```diff
@@ -0,0 +1,122 @@
...
+  Include:
+    - 'src/lib/**/*.rb' # force well documented code only for new code
+
+Style/FileName:
+  Include:
+    - 'src/lib/**/*.rb' # force correct names on new code
+
+Style/HashSyntax:
+  Include:
+    - 'src/lib/**/*.rb' # force new hash style only for new code
```
jreidinger: why? if you do adaptation, then hashSyntax can be unified automatic.

test/REPORT
teclator: Ahahah , big mistake

.rubocop.yml
teclator: Yeah, i just looked for some yast module that have it updated so i copy it from yast-bootloader module :wink:

src/include/ntp-client/dialogs.rb
teclator: not rubocop 0.29.1 at least

.rubocop.yml
imobach: typo: `then` -> `than`

src/include/ntp-client/dialogs.rb
jreidinger: reason is that you disable it in config, as I noticed and commented now

src/modules/NtpClient.rb
teclator: So this commend can be dropped also.

src/include/ntp-client/helps.rb
```diff
@@ -28,45 +28,45 @@ def initialize_ntp_client_helps(include_target)
...
           ),
         # help text 1/5, %d is a number of minutes
         "start"              => _(
-          "<p><b><big>Start NTP Daemon</big></b><br>\n" +
-            "Select whether to start the NTP daemon now and on every system boot. \n" +
-	    "Selecting <b>Synchronize without Daemon</b> the NTP daemon will not be activated\n" +
-	    "and the system time will be set periodically by a <i>cron</i> script. \n" +
+          "<p><b><big>Start NTP Daemon</big></b><br>\n" \
+            "Select whether to start the NTP daemon now and on every system boot. \n" \
+      "Selecting <b>Synchronize without Daemon</b> the NTP daemon will not be activated\n" \
```
imobach: Is indentation correct?

src/include/ntp-client/helps.rb
```diff
@@ -28,45 +28,45 @@ def initialize_ntp_client_helps(include_target)
...
+      "and the system time will be set periodically by a <i>cron</i> script. \n" \
       "The interval is configurable, by default it is %d minutes."
-	    ) % NtpClientClass::DEFAULT_SYNC_INTERVAL,
+      ) % NtpClientClass::DEFAULT_SYNC_INTERVAL,
         # help text 2/5
         "chroot_environment" => _(
-          "<p><b><big>Chroot Jail</big></b><br>\n" +
-            "To run the NTP daemon in chroot jail, set\n" +
-            "<b>Run NTP Daemon in Chroot Jail</b>. Starting any daemon in a chroot jail\n" +
+          "<p><b><big>Chroot Jail</big></b><br>\n" \
```
imobach: Is indentation correct?

src/include/ntp-client/misc.rb
```diff
@@ -113,7 +111,7 @@ def string2opts(options_string, with_param, without_param)
           next { e => nil }
         end
       end
-      ret = Builtins.filter(ret) { |k, v| v != nil }
+      ret = Builtins.filter(ret) { |_k, v| !v.nil? }
```
imobach: You could also use "#reject" here.

src/include/ntp-client/widgets.rb
```diff
@@ -174,16 +170,14 @@ def startHandle(id, event)
...
-          !# continue-cancel popup
-          Popup.ContinueCancel(
+          Ops.get_string(event, "EventReason", "") != "" && !Popup.ContinueCancel(
             _(
-              "Warning!\n" +
-                "\n" +
-                "If you do not have a permanent Internet connection,\n" +
-                "starting the NTP daemon can take a very long time and \n" +
+              "Warning!\n" \
+                "\n" \
```
imobach: Indentation is not consistente (check line 179-181).

src/include/ntp-client/widgets.rb
```diff
@@ -305,7 +300,7 @@ def complexButtonHandle(key, event)
         # disabled in case of PoolNTPorg feature
         # and in case of missing server value
         if Convert.to_boolean(UI.QueryWidget(Id("server_address"), :Enabled)) &&
-            Ops.get(NtpClient.selected_record, "address") != nil &&
+            !NtpClient.selected_record["address"].nil? &&
```
imobach: np: Maybe you could just do `NtpClient.selected_record["address"]`...

src/include/ntp-client/widgets.rb
```diff
@@ -531,12 +518,12 @@ def RandomServersHandle(id, event)
...
-              _(
-                "Enabling Random Servers from pool.ntp.org would\n" +
-                  "replace the current NTP server.\n" +
-                  "\n" +
-                  "Really replace the current NTP server?"
-              )
+            _(
+              "Enabling Random Servers from pool.ntp.org would\n" \
+              "replace the current NTP server.\n"                 \
+              "\n"                                                \
```
imobach: np: Why not must move this `\n` to the line above?

src/include/ntp-client/dialogs.rb
teclator: To quick answer, my assumption was absolutely wrong, as also all rules are inherit from /usr/share/YaST2/data/devtools/data/rubocop_yast_style.yml, yeah, some rules are only applied for new code under src/lib

src/include/ntp-client/widgets.rb
```diff
@@ -1062,13 +1042,12 @@ def unitNumberStore(id, event)
...
     # @return [Symbol] always nil
-    def browseButtonHandle(id, event)
-      event = deep_copy(event)
+    def browseButtonHandle(_id, _event)
       current = Convert.to_string(UI.QueryWidget(Id("device"), :Value))
       current = "/dev" if current == ""
       # popup header
       current = UI.AskForExistingFile(current, "", _("Select the Device"))
-      UI.ChangeWidget(Id("device"), :Value, current) if current != nil
+      UI.ChangeWidget(Id("device"), :Value, current) if !current.nil?
```
imobach: Or you could use `unless current.nil?`

src/include/ntp-client/widgets.rb
```diff
@@ -1186,28 +1162,27 @@ def ServerListInit(id)
       @last_country = nil
 
       lang = NtpClient.GetCurrentLanguageCode
-      if lang != nil
+      if !lang.nil?
```
imobach: np: if `lang`

src/include/ntp-client/widgets.rb
```diff
@@ -1463,47 +1430,47 @@ def InitWidgets
...
-              [:custom, _("Custom")]
-            ],
-            "label"  => _("&Runtime Configuration Policy"),
-            "init"   => fun_ref(method(:PolicyInit), "void (string)"),
-            "store"  => fun_ref(method(:PolicyStore), "void (string, map)"),
-            "handle" => fun_ref(
-              method(:ntpEnabledOrDisabled),
-              "symbol (string, map)"
+        "policy_combo"       =>
+                                # FIXME
```
imobach: It looks like too much indentation..

src/modules/NtpClient.rb
```diff
@@ -150,7 +147,7 @@ def PolicyIsNonstatic
     # Abort function
     # @return blah blah lahjk
     def Abort
-      return @AbortFunction.call == true if @AbortFunction != nil
+      return @AbortFunction.call == true if !@AbortFunction.nil?
```
imobach: I would use `unless` here.

imobach: What about this one-liner?

```
@AbortFunction.nil? ? false : @AbortFunction.call == true
```

src/modules/NtpClient.rb
```diff
@@ -623,13 +586,8 @@ def Read
...
     def GetUsedNtpServers
       used_servers = []
-      Builtins.foreach(@ntp_records) do |record|
-        if Ops.get_string(record, "type", "") == "server"
-          used_servers = Builtins.add(
-            used_servers,
-            Ops.get_string(record, "address", "")
-          )
-        end
+      @ntp_records.each do |record|
```
imobach: Maybe you would like to use `each_with_object` here.

src/modules/NtpClient.rb
```diff
@@ -267,7 +245,7 @@ def GetNtpServersByCountry(country, terse_output)
       country_names = {}
       servers = GetNtpServers()
       if country != ""
-        servers = Builtins.filter(servers) do |s, o|
+        servers = Builtins.filter(servers) do |_s, o|
```
imobach: `s` and `o` would deserve better names :)

src/modules/NtpClient.rb
```diff
@@ -780,70 +736,63 @@ def Write
       return false if Abort()
       Progress.NextStage if have_progress
 
-      if true
```
imobach: Ouch

src/modules/NtpClient.rb
```diff
@@ -780,70 +736,63 @@ def Write
...
-
-        Report.Error(Message.CannotWriteSettingsTo("/etc/ntp.conf")) if failed
+      failed = false
+      conf = Convert.to_map(SCR.Read(path(".etc.ntp_conf.all")))
+      if conf.nil?
+        failed = true
+      else
+        Ops.set(conf, "value", save2)
+        failed = true if !SCR.Write(path(".etc.ntp_conf.all"), conf)
+        failed = true if !SCR.Write(path(".etc.ntp_conf"), nil)
```
imobach: One condition should be enough.

src/include/ntp-client/misc.rb
teclator: And also #select but if I change it here then i should clean up the whole method

test/ntp_client_test.rb
```diff
@@ -0,0 +1,487 @@
...
+Yast.import "Service"
+
+describe Yast::NtpClient do
+
+  subject { Yast::NtpClient }
+
+  describe "#Read" do
+    before do
+      subject.config_has_been_read = false
+      allow(Yast::NetworkInterfaces).to receive(:Read)
```
imobach: Oh, too many allows :)

test/ntp_client_test.rb
```diff
@@ -0,0 +1,487 @@
...
+      allow(subject).to receive(:Abort).and_return(false)
+      allow(Yast::Mode).to receive(:normal).and_return(false)
+      allow(Yast::Progress).to receive(:set)
+      allow(Yast::SCR).to receive(:Read)
+        .with(Yast::Path.new(".sysconfig.network.config.NETCONFIG_NTP_POLICY"))
+        .and_return(nil)
+      allow(Yast::PackageSystem).to receive(:CheckAndInstallPackagesInteractive)
```
imobach: I would avoid mocking with `allows` those method calls that are not used (almost) always. I prefer using `context` for that.

test/ntp_client_test.rb
```diff
@@ -0,0 +1,487 @@
...
+describe Yast::NtpClient do
+
+  subject { Yast::NtpClient }
+
+  describe "#Read" do
+    before do
+      subject.config_has_been_read = false
+      allow(Yast::NetworkInterfaces).to receive(:Read)
+      allow(subject).to receive(:Abort).and_return(false)
+      allow(Yast::Mode).to receive(:normal).and_return(false)
```
imobach: You could mock `Yast::Mode.mode` so we you'll get `Yast::Mode.normal`, `Yast::Mode.installation`, etc. for free :) BTW, is the `Mode.install == true` scenario tested?

test/ntp_client_test.rb
```diff
@@ -0,0 +1,487 @@
...
+
+    it "loads known ntp servers and known country names" do
+      expect(subject).to receive(:GetNtpServers)
+      expect(subject).to receive(:GetCountryNames)
+
+      subject.Read
+    end
+
+    it "returns false if the ntp package is not and could not be installed" do
+      expect(Yast::PackageSystem).to receive(:CheckAndInstallPackagesInteractive)
```
imobach: Forget my previous comment about `CheckAndInstallPackagesInteractive` :)

test/ntp_client_test.rb
```diff
@@ -0,0 +1,487 @@
...
+    it "returns true if all reads were performed" do
+      expect(subject.Read).to eql(true)
+    end
+  end
+
+  describe "#Write" do
+    let(:data_dir) { File.join(File.dirname(__FILE__), "data") }
+
+    around do |example|
+      ::FileUtils.cp(File.join(data_dir, "scr_root/etc/ntp.conf.original"),
```
imobach: Maybe it's just a matter of taste, but I prefer to not modify files which are under version control. I would copy the whole directory to a temporal one to perform tests. The reason is that you could modify those files during a test run and commit them by accident.

And what if you have three or four configuration files?

test/ntp_client_test.rb
```diff
@@ -0,0 +1,487 @@
...
+      change_scr_root(File.join(data_dir, "scr_root"), &example)
+    end
+
+    before do
+      allow(Yast::Mode).to receive(:normal).and_return(false)
+      allow(subject).to receive(:Abort).and_return(false)
+      allow(Yast::SCR).to receive(:Read)
+        .with(Yast::Path.new(".etc.ntp_conf.all")).and_return(nil)
+      allow(Yast::FileChanges).to receive(:StoreFileCheckSum).with("/etc/ntp.conf")
+      allow(Yast::SCR).to receive(:Write)
```
imobach: Just asking: do you need to mock this call? You're using a `scr_root/etc/sysconfig/network` directory, right?

test/ntp_client_test.rb
```diff
@@ -0,0 +1,487 @@
...
+        .with(Yast::Path.new(".sysconfig.network.config.NETCONFIG_NTP_POLICY"), anything)
+      allow(Yast::SCR).to receive(:Write)
+        .with(Yast::Path.new(".sysconfig.network.config"), nil)
+      allow(Yast::SCR).to receive(:Execute)
+        .with(Yast::Path.new(".target.bash"), "/sbin/netconfig update -m ntp")
+        .and_return(0)
+      allow(Yast::SCR).to receive(:Write)
+        .with(Yast::Path.new(".sysconfig.ntp.NTPD_RUN_CHROOTED"), anything)
```
imobach: same as above.

test/ntp_client_test.rb
```diff
@@ -0,0 +1,487 @@
...
+          "value"   => []
+        }
+      end
+
+      before do
+        allow(Yast::SCR).to receive(:Read)
+          .with(Yast::Path.new(".etc.ntp_conf.all")).and_call_original
+      end
+
+      it "tries to write to it current ntp entries" do
```
imobach: I would omit `to it`.

test/ntp_client_test.rb
```diff
@@ -0,0 +1,487 @@
...
+        expect(Yast::SCR).to receive(:Write).with(Yast::Path.new(".etc.ntp_conf.all"), conf)
+        expect(Yast::SCR).to receive(:Write).with(Yast::Path.new(".etc.ntp_conf"), nil)
+
+        subject.Write
+      end
+
+      it "reports and error if not able to write /etc/ntp.conf file" do
```
imobach: _an error_ and _if it's not able_

test/ntp_client_test.rb
```diff
@@ -0,0 +1,487 @@
...
+        .and_return(0)
+      allow(Yast::SCR).to receive(:Write)
+        .with(Yast::Path.new(".sysconfig.ntp.NTPD_RUN_CHROOTED"), anything)
+      allow(Yast::SCR).to receive(:Write)
+        .with(Yast::Path.new(".sysconfig.ntp"), nil)
+      allow(Yast::SCR).to receive(:Execute)
+        .with(Yast::Path.new(".target.bash"),
+          "test -e #{subject.cron_file} && rm #{subject.cron_file};")
+      allow(Yast::SuSEFirewall).to receive(:Write)
+      allow(Yast::Service).to receive(:Enable)
```
imobach: All of them are called all the time? Maybe we could think of a better organization. But with such a long method...

test/ntp_client_test.rb
```diff
@@ -0,0 +1,487 @@
...
+    context "when ntp servers haven't been read before" do
+      it "returns empty hash in test Mode" do
+        allow(Yast::Mode).to receive(:test).and_return(true)
+
+        expect(subject.GetNtpServers).to eql({})
+      end
+
+      it "caches known ntp servers" do
+        allow(Yast::Mode).to receive(:test).and_return(false)
+        subject.instance_variable_set(:@ntp_servers, nil)
```
imobach: Maybe there's a way of resetting `@ntp_servers` without using `instance_variable_set`. Maybe calling `main`?

test/ntp_client_test.rb
```diff
@@ -0,0 +1,487 @@
...
+        expect(subject).to receive(:update_ntp_servers!)
+
+        subject.GetNtpServers
+      end
+
+      it "return known ntp servers" do
+        subject.instance_variable_set(:@ntp_servers, country_servers)
```
imobach: You shouldn't rely on instance variables. You're loading NTP servers from a YAML file, right? So you can check that those servers are returned.

test/ntp_client_test.rb
```diff
@@ -0,0 +1,487 @@
...
+
+      it "caches known ntp servers" do
+        allow(Yast::Mode).to receive(:test).and_return(false)
+        subject.instance_variable_set(:@ntp_servers, nil)
+        expect(subject).to receive(:update_ntp_servers!)
+
+        subject.GetNtpServers
+      end
+
+      it "return known ntp servers" do
```
imobach: _returns_

test/ntp_client_test.rb
```diff
@@ -0,0 +1,487 @@
...
+        subject.instance_variable_set(:@ntp_servers, country_servers)
+
+        expect(subject.GetNtpServers).to eql(country_servers)
+      end
+    end
+
+    context "when ntp servers haven been read before" do
```
imobach: _have_

test/ntp_client_test.rb
```diff
@@ -0,0 +1,487 @@
...
+
+  end
+
+  describe "#reachable_ntp_server?" do
+    context "given a server" do
+      it "returns true if sntp test passed with IPv4 or/and with IPv6" do
+        allow(subject).to receive(:sntp_test).with("server").and_return(true, false, true)
+        allow(subject).to receive(:sntp_test).with("server", 6).and_return(true, true)
+
+        expect(subject.reachable_ntp_server?("server")).to eql(true)
```
imobach: Why are you calling it three times? Wouldn't be better to have different tests?

test/ntp_client_test.rb
```diff
@@ -0,0 +1,487 @@
...
+    end
+
+  end
+
+  describe "#sntp_test" do
+    let(:ip_version) { 4 }
+    let(:server) { "sntp.server.de" }
+    let(:output) { { "stdout" => "", "stderr" => "", "exit" => 0 } }
+
+    context "given a server" do
```
imobach: Maybe this context is not needed, as you don't have a `not given a server` one :)

src/include/ntp-client/helps.rb
```diff
@@ -235,14 +245,15 @@ def initialize_ntp_client_helps(include_target)
     def fudgeHelp
       # help text 1/2
       _(
-        "<p><big><b>Clock Driver Calibration</b></big><br>\n" +
-          "The clock driver may need to be calibrated. In this dialog, various calibration\n" +
-          "options can be set. The meaning of particular options depends on the particular\n" +
+        "<p><big><b>Clock Driver Calibration</b></big><br>\n" \
+          "The clock driver may need to be calibrated. In this dialog, various calibration\n" \
```
imobach: Please, check indentantion.

src/include/ntp-client/misc.rb
```diff
@@ -39,7 +39,7 @@ def reallyExitSimple
       if NtpClient.run_service ==
           (UI.QueryWidget(Id("start"), :CurrentButton) == "boot") &&
           Ops.get_string(NtpClient.selected_record, "address", "") ==
-            UI.QueryWidget(Id("server_address"), :Value)
+              UI.QueryWidget(Id("server_address"), :Value)
```
imobach: np: Is identation ok? Not sure.

src/include/ntp-client/misc.rb
```diff
@@ -82,40 +81,32 @@ def silentWrite
...
+      parsed = {}
       unknown = []
-      ret = Builtins.listmap(l) do |e|
-        index = Ops.add(index, 1)
-        if ignore_next
-          ignore_next = false
-          next { e => nil }
+      options_list.each do |option|
+        if option_parse
+          parsed[option_parse] = option unless option.nil?
```
imobach: How can it be `option.nil?`?

src/modules/NtpClient.rb
```diff
@@ -266,92 +253,59 @@ def GetCountryNames
...
-        if label != "" && l_country != ""
-          label = Builtins.sformat("%1 (%2, %3)", s, label, l_country)
-        elsif label == "" && l_country == ""
-          label = s
-        else
-          label = Builtins.sformat("%1 (%2%3)", s, label, l_country)
-        end
-
+      default = false
+      items = servers.map do |server, attrs|
```
imobach: If this `map` is the last call of the method, you can avoid using an `items` variable, right?

src/modules/NtpClient.rb
```diff
@@ -266,92 +253,59 @@ def GetCountryNames
...
-              "broadcast",
-              "broadcastclient",
-              "manycast",
-              "manycastclient",
-              "fudge",
-              "restrict"
-            ],
-            type
-          )
+        if ["server", "peer", "broadcast", "broadcastclient", "manycast",
```
imobach: Maybe you could define a constant before the method (are at least add a comment about what are these terms).

src/modules/NtpClient.rb
```diff
@@ -648,13 +525,13 @@ def IsRandomServersServiceEnabled
...
         # if server is needed by pool.ntp.org and matches
-        if Ops.get(needed_servers, used_server) != nil
+        if !Ops.get(needed_servers, used_server).nil?
           Ops.set(needed_servers, used_server, true)
         end
       end
 
       ret = true
-      Builtins.foreach(needed_servers) do |nserver_name, ns_value|
+      Builtins.foreach(needed_servers) do |_nserver_name, ns_value|
```
imobach: Maybe you could use `all?`. But I guess it's out of the scope of this PR :)

src/modules/NtpClient.rb
```diff
@@ -1263,56 +998,294 @@ def AutoPackages
...
+        options = " "
+        options << "mask #{m["mask"]} " if !m["mask"].to_s.empty?
+        options << m["options"].to_s
+        ret = {
+          "address" => key,
+          "comment" => m["comment"].to_s,
+          "type"    => "restrict",
+          "options" => options
+        }
+        records << ret
```
imobach: You're modifying an argument. I would prefer one of these options:

1) Don't modify the argument: `@ntp_records += restrict_records`
2) Modify `@ntp_records` directly (after all, it's a instance variable).

src/modules/NtpClient.rb
```diff
@@ -1263,56 +998,294 @@ def AutoPackages
...
+          # progress step
+          _("Restarting NTP daemon..."),
+          # progress step
+          _("Finished")
+        ],
+        ""
+      )
+    end
+
+    def write_ntp_conf
```
imobach: Please, document the method.

src/modules/NtpClient.rb
```diff
@@ -1263,56 +998,294 @@ def AutoPackages
...
+      @ntp_servers = {}
+
+      read_known_servers.each { |s| cache_server(s) }
+
+      pool_servers_for(GetAllKnownCountries()).each { |p| cache_server(p) }
+    end
+
+  private
+
+    def new_read_progress(caption, steps)
```
imobach: Please, document the method.

src/modules/NtpClient.rb
```diff
@@ -1263,56 +998,294 @@ def AutoPackages
...
+    def write_chroot_config
+      SCR.Write(
+        path(".sysconfig.ntp.NTPD_RUN_CHROOTED"),
+        @run_chroot ? "yes" : "no"
+      )
+
+      SCR.Write(path(".sysconfig.ntp"), nil)
+    end
+
+    def update_netconfig
```
imobach: Please, document the method. Same below.

src/modules/NtpClient.rb
```diff
@@ -740,193 +616,41 @@ def ActivateRandomPoolServersFunction
...
-      end
+      write_ntp_policy && update_netconfig
 
-      SCR.Write(
-        path(".sysconfig.ntp.NTPD_RUN_CHROOTED"),
-        @run_chroot ? "yes" : "no"
-      )
-      SCR.Write(path(".sysconfig.ntp"), nil)
-
-      Builtins.sleep(sl)
```
imobach: Is the sleep not needed anymore?

src/modules/NtpClient.rb
```diff
@@ -1263,56 +998,294 @@ def AutoPackages
...
+    end
+
+    def check_service
+      adjusted = @run_service ? Service.Enable(@service_name) : Service.Disable(@service_name)
+
+      # error report
+      Report.Error(Message.CannotAdjustService("NTP")) unless adjusted
+
+      if @run_service
+        unless @write_only
```
imobach: Why not combine this condition with the one in the next line?

test/ntp_client_test.rb
```diff
@@ -0,0 +1,618 @@
...
+    let(:data_dir) { File.join(File.dirname(__FILE__), "data") }
+
+    around do |example|
+      change_scr_root(File.join(data_dir, "scr_root_read"), &example)
+    end
+
+    before do
+      allow(subject).to receive(:Abort).and_return(false)
+      allow(Yast::NetworkInterfaces).to receive(:Read)
+      allow(Yast::Mode).to receive(:normal).and_return(false)
```
imobach: To set the mode, you could mock `Yast::Mode.mode`.

test/ntp_client_test.rb
```diff
@@ -0,0 +1,618 @@
...
+        subject.Read
+      end
+
+      it "reads network interfaces config" do
+        expect(Yast::NetworkInterfaces).to receive(:Read)
+
+        subject.Read
+      end
+
+      context "when sysconfig network policity setting's config exists" do
```
imobach: policy? I don't get the whole sentence anyway (maybe it's just me).

test/ntp_client_test.rb
```diff
@@ -0,0 +1,618 @@
...
+        it "sets ntp policy setting as read value" do
+          allow(Yast::FileChanges).to receive(:CheckFiles).with(["/etc/ntp.conf"])
+          allow(Yast::SCR).to receive(:Read)
+            .with(Yast::Path.new(".sysconfig.network.config.NETCONFIG_NTP_POLICY"))
+            .and_return("manual")
```
imobach: I would use something like `some-policy`. Maybe it's set to `manual` in another piece of code and this test will keep passing.

test/ntp_client_test.rb
```diff
@@ -0,0 +1,618 @@
...
+
+          subject.Read
+
+          expect(subject.ntp_policy).to eql("manual")
+        end
+      end
+
+      context "when sysconfig network policity setting's config doesn't exist" do
```
imobach: Same as above.

test/ntp_client_test.rb
```diff
@@ -0,0 +1,618 @@
...
+
+      it "loads known ntp servers and known country names" do
+        expect(subject).to receive(:GetNtpServers)
+        expect(subject).to receive(:GetCountryNames)
+
+        subject.Read
+      end
+
+      context "when Mode is not installation" do
+        it "returns false if the ntp package is not and could not be installed" do
```
imobach: _neither is installed nor available_?

src/modules/NtpClient.rb
```diff
@@ -1263,56 +998,294 @@ def AutoPackages
...
-    publish :function => :TestNtpServer, :type => "boolean (string, symbol)"
-    publish :function => :DetectNtpServers, :type => "list <string> (symbol)"
-    publish :function => :getSyncRecords, :type => "list <map <string, any>> ()"
-    publish :function => :selectSyncRecord, :type => "boolean (integer)"
-    publish :function => :findSyncRecord, :type => "integer (string, string)"
-    publish :function => :storeSyncRecord, :type => "boolean ()"
-    publish :function => :deleteSyncRecord, :type => "boolean (integer)"
-    publish :function => :enableOptionInSyncRecord, :type => "void (string)"
-    publish :function => :AutoPackages, :type => "map ()"
+    def read_ad_address!
```
imobach: Is this method really public?

test/ntp_client_test.rb
```diff
@@ -0,0 +1,618 @@
...
+    end
+
+    it "returns false if abort is pressed" do
+      allow(subject).to receive(:Abort).and_return(true)
+      expect(subject).to receive(:go_next).and_call_original
+
+      expect(subject.Write).to eql(false)
+    end
+
+    it "appends restrict records to current ntp_records before write" do
```
imobach: writing

test/ntp_client_test.rb
```diff
@@ -0,0 +1,618 @@
...
+
+    it "gets all country names if given country name is an empty string" do
+      expect(subject).to receive(:GetNtpServers).and_call_original
+      expect(subject).to receive(:GetCountryNames).and_call_original
+
+      subject.GetNtpServersByCountry("", false)
+    end
+
+    it "returns a list of items with read servers" do
+
```
imobach: Pending?

src/modules/NtpClient.rb
teclator: A sleep of 0 makes sense?

src/modules/NtpClient.rb
teclator: I was thinking on move this to a constant, and also some of the variables as for example the cron entry or cron file for synchronize_time

src/modules/NtpClient.rb
teclator: Some of the changes are just for make rubocop happier. There is a lot of methods that could be rewrite but I’m trying to add tests before touch or clean some method.

test/ntp_client_test.rb
```diff
@@ -0,0 +1,618 @@
...
+    it "returns true if ntp server is reachable" do
+      allow(subject).to receive(:reachable_ntp_server?).with("server") { true }
+      expect(subject.TestNtpServer("server", "")).to eql(true)
+    end
+    it "returns false if ntp server is not reachable" do
+      allow(subject).to receive(:reachable_ntp_server?).with("server") { false }
+      expect(subject.TestNtpServer("server", "")).to eql(false)
+    end
+
+    context "when given verbosity is no ui" do
```
imobach: I would expect something setting the _verbosity_ to `no ui`.

test/ntp_client_test.rb
```diff
@@ -0,0 +1,618 @@
...
+        expect(Yast::Popup).to receive(:Feedback).never
+      end
+    end
+  end
+
+  describe "#getSyncRecords" do
+    let(:data_dir) { File.join(File.dirname(__FILE__), "data") }
+
+    around do |example|
+      ::FileUtils.cp(File.join(data_dir, "scr_root/etc/ntp.conf.original"),
```
imobach: You should remove `scr_root/etc/ntp.conf` at the end of the test.

imobach:
Nice clean-up! Much better than before. I've added some minor comments anyway.

src/modules/NtpClient.rb
imobach: Ok, it's fine with me :)

src/modules/NtpClient.rb
```diff
@@ -21,6 +21,18 @@ class NtpClientClass < Module
     # Note: the UI field currently uses maximum of 60 minutes
     DEFAULT_SYNC_INTERVAL = 5
 
+    # List of servers defined by the pool.ntp.org to get random ntp servers
+    #
+    # @see #http://www.pool.ntp.org/
+    RANDOM_POOL_NTP_SERVERS = ["0.pool.ntp.org", "1.pool.ntp.org", "2.pool.ntp.org"]
+
+    # Different kinds of records wich the server can syncronize with and
```
imobach: `which`

src/modules/NtpClient.rb
```diff
@@ -21,6 +21,18 @@ class NtpClientClass < Module
...
+    # reference clock record
+    #
+    # @see http://doc.ntp.org/4.1.0/confopt.htm
+    # @see http://doc.ntp.org/4.1.0/clockopt.htm
+    SYNC_RECORDS = ["server", "__clock", "peer", "broadcast", "broadcastclient"]
```
imobach: :+1:

src/modules/NtpClient.rb
```diff
@@ -509,48 +509,24 @@ def GetUsedNtpServers
...
-          Ops.set(needed_servers, used_server, true)
-        end
-      end
-
-      ret = true
-      Builtins.foreach(needed_servers) do |_nserver_name, ns_value|
-        ret = false if ns_value != true
-      end
-      ret
+      RANDOM_POOL_NTP_SERVERS.all? { |server| used_servers.include? server }
```
imobach: np: According to our style guide, you should use a 1-letter identifier when using `{}` blocks.

src/modules/NtpClient.rb
```diff
@@ -509,48 +509,24 @@ def GetUsedNtpServers
...
-            # do not filter out not-servers
-            one_record,
-            "type",
-            ""
-          ) != "server" ||
-            # do not filter out serces that are not random_pool_servers
-            Ops.get_string(one_record, "address", "") != random_pool_server
-        end
+      @ntp_records.reject! do |record|
+        record["type"] == "server" && RANDOM_POOL_NTP_SERVERS.include?(record["address"])
```
imobach: Much more readable. Thanks.

src/modules/NtpClient.rb
```diff
@@ -779,24 +755,12 @@ def Summary
...
-          summary = Summary.AddLine(
-            summary,
-            Builtins.sformat(
-              Ops.get_string(types, t, ""),
-              Builtins.mergestring(names, ", ")
-            )
-          )
-        end
+      SYNC_RECORDS.each do |t|
+        type_records = @ntp_records.select { |record| record["type"] == t }
```
imobach: np: 1-char variables for these blocks.

src/modules/NtpClient.rb
```diff
@@ -918,20 +882,12 @@ def getSyncRecords
     # @return [Boolean] true on success
     def selectSyncRecord(index)
       ret = true
-      if Ops.greater_or_equal(index, Builtins.size(@ntp_records)) ||
-          Ops.less_than(index, -1)
-        Builtins.y2error(
-          "Record with index %1 doesn't exist, creating new",
-          index
-        )
+      if index >= @ntp_records.size || index < -1
```
imobach: I would prefer something like `(-1..index).cover?(@ntp_records.size)`.

src/modules/NtpClient.rb
```diff
@@ -998,6 +954,16 @@ def AutoPackages
       { "install" => @required_packages, "remove" => [] }
     end
 
+  private
+
+    # It reads ntp policy from sysconfig, sets as auto is not able to read it.
```
imobach: `if it's not able`

src/modules/NtpClient.rb
```diff
@@ -1018,6 +984,9 @@ def read_ad_address!
       end
     end
 
+    # Reads /etc/sysconfig/ntp NTPD_RUN_CHROOTED and sets @run_chroot as true
```
imobach: This comment is a matter of personal taste, so take it with a grain of salt. When documenting, I prefer to _summarize_ in the first line what the method does, clarifying below if needed. In this case, you could also write something like:

```
# Set @run_chroot according to NTPD_RUN_CHROOTED value found in /etc/sysconfig/ntp
#
# @return [Boolean] true when value is "yes"; false in any other case.
```

src/modules/NtpClient.rb
```diff
@@ -1118,14 +1112,9 @@ def write_chroot_config
...
 
-    def update_netconfig
-      error = SCR.Execute(path(".target.bash"), "/sbin/netconfig update -m ntp") != 0
-
-      Report.Error(_("Cannot update the dynamic configuration policy.")) if error
-
-      !error
-    end
-
+    # Enable or disable ntp service depending on @run_service value. In case of
```
imobach: I would prefer:

```
# Enable or disable ntp service depending on @run_service value
#
# * When disabling, it also stops the service.
# * When enabling, it tries to restart the service unless it's in write only mode.
```

Please, feel free to ignore this suggestion as it's a matter of personal preference.

src/modules/NtpClient.rb
```diff
@@ -1177,28 +1170,34 @@ def fugde_options_to_h(record)
...
           "comment" => m["comment"].to_s,
           "type"    => "restrict",
           "options" => options
         }
-        records << ret
       end
     end
 
+    # Prepare current ntp_records in hashes for write, splitting fudge options
+    # of __clock records in their own hashes.
```
imobach: Please, document param `records`.

src/modules/NtpClient.rb
```diff
@@ -1215,23 +1214,47 @@ def read_known_servers
       servers
     end
 
-    def country_server_label(location, country)
+    # Returns a concatenation of given location and country depending on if
+    # them are empty or not.
+    #
+    # ==== Examples
```
imobach: You should use [@example tag](http://www.rubydoc.info/gems/yard/file/docs/Tags.md#example) here.

test/ntp_client_test.rb
```diff
@@ -347,18 +368,23 @@
 
   describe "#reachable_ntp_server?" do
     context "given a server" do
-      it "returns true if sntp test passed with IPv4 or/and with IPv6" do
-        allow(subject).to receive(:sntp_test).with("server").and_return(true, false, true)
-        allow(subject).to receive(:sntp_test).with("server", 6).and_return(true, true)
+      it "returns true if sntp test passed with IPv4" do
```
imobach: Much better now.

test/ntp_client_test.rb
```diff
@@ -421,41 +447,150 @@
       expect(subject.TestNtpServer("server", "")).to eql(false)
     end
 
-    context "when given verbosity is no ui" do
-      it "doesn't show any dialog" do
-        expect(Yast::Popup).to receive(:Feedback).never
-      end
+    it "doesn't show any dialog if given verbosity is :no_ui" do
```
imobach: Are you checking that some pop-up is shown when verbosity is not `:no_ui`?

test/ntp_client_test.rb
```diff
@@ -421,41 +447,150 @@
...
+  describe "#selectSyncRecord" do
+    let(:data_dir) { File.join(File.dirname(__FILE__), "data") }
+
+    around do |example|
+      change_scr_root(File.join(data_dir, "scr_root_read"), &example)
+    end
+
+    context "when given index is not between -1 an ntp_records size" do
+      it "returns false" do
+        subject.instance_variable_set(:@config_has_been_read, false)
```
imobach: Maybe you could move the first two lines of each scenario to a `before` block.

test/ntp_client_test.rb
teclator: no, just the no_ui case to return directly reachable_ntp_server?, but i can i can cover this cases also.

src/modules/NtpClient.rb
```diff
@@ -925,15 +925,12 @@ def storeSyncRecord
     # @param [Fixnum] index integer index of record to delete
     # @return [Boolean] true on success
     def deleteSyncRecord(index)
-      if Ops.greater_or_equal(index, Builtins.size(@ntp_records)) ||
-          Ops.less_or_equal(index, -1)
-        Builtins.y2error("Record with index %1 doesn't exist", index)
+      if index >= @ntp_records.size || index <= -1
```
imobach: Same as `selectSyncRecord` (use `#cover?`).

imobach:
LGTM

PR #60, opened 2016-06-14
Convert ntp record's comment to string to avoid nil values (bsc#977610)

teclator:
If nil comments are given then an empty server entry will be written to /etc/ntp.conf


imobach:
LGTM

PR #61, opened 2016-06-30
Fix empty cron (bsc#987064)

jreidinger:


package/yast2-ntp-client.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Jun 30 06:51:52 UTC 2016 - jreidinger@suse.com
+
+- Fix regression: crash when reading cron entry failed
+  ( discovered by sap-ha module ) (bsc#977610)
```
mvidner: The bug only describes the original problem, not the one that we are fixing now.
Please add a description of the new problem, otherwise we risk making a wrong fix.

jreidinger: It is described: crash when reading cron entry failed - in this case SCR agent return nil and code crash

jreidinger: I do not have new bug report for it. It failed in build service

mvidner: Thanks for adding a description to the bug!

mvidner:
The CI fails (because of https://github.com/yast/yast-ruby-bindings/pull/169) but I am confident the fix is good.

LGTM

teclator:
@jreidinger thanx for the fix, it would be fine to also cover this case with a test

test/ntp_client_test.rb
```diff
@@ -351,17 +353,33 @@
...
+          expect(subject.sync_interval).to eql(Yast::NtpClientClass::DEFAULT_SYNC_INTERVAL)
+        end
       end
 
-      it "sets sync_interval with cron minute interval" do
-        subject.ReadSynchronization
+      context "when there is cron entry" do
+        let(:cron_entry) { [{ "events"   => [{ "active"   => "1", "minute" => "*/10" }] }] }
+
+        it "sets synchronize time as true if first cron entry is valid" do
```
imobachgs: np: "returns true..." (feel free to ignore) :)

imobachgs:
LGTM

PR #63, opened 2016-10-18
Fix check

jreidinger:
even if previous comment that it is too dependent in future, I think customer interested is more important and when we provide "Test" functionality it should work reliable.

### DIFF
--- a/src/modules/NtpClient.rb
+++ b/src/modules/NtpClient.rb
@@ -793,8 +793,8 @@ module Yast
       # so this is a workaround at least to return false in case server is not
       # reachable.
       return false if output["stderr"].include?("lookup error")
-      # this check is for valid address, but without ntp. If it break in future
-      # start complaining to sntp maintainer to not return 0 in this case.
+      # this happens for valid address, but without ntp server. If it breaks in the
+      # future start complaining to sntp maintainer to not return 0 in this case.
       # customer case: bsc#972842
       return false if output["stdout"] =~ /no (U|B)CST/
 
--- a/src/modules/NtpClient.rb
+++ b/src/modules/NtpClient.rb
@@ -788,13 +788,17 @@ module Yast
 
       Builtins.y2milestone("sntp test response: #{output}")
 
+      return false if output["exit"] != 0
       # sntp returns always 0 if not called with option -S or -s (set system time)
       # so this is a workaround at least to return false in case server is not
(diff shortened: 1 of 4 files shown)


mvidner:
Unrelated to this fix, but `LANG=en_US rspec -f d test/ntp_client_test.rb` fails with "uninitialized constant FileUtils", please `require "fileutils"`.

test/ntp_client_test.rb
```diff
@@ -442,6 +442,17 @@
           expect(subject.sntp_test(server)).to eql(false)
         end
 
+        it "returns false if sntp respong no UCST" do
+          output["stdout"] = "sntp 4.2.8p8@1.3265-o Fri Sep 30 15:52:10 UTC 2016 (1)\n" \
+            "195.113.144.2 no UCST response after 5 seconds\n"
```
mvidner: `output["stderr"] =""`, otherwise it would mean the test passes even if you got the regex wrong.

test/ntp_client_test.rb
```diff
@@ -442,6 +442,17 @@
           expect(subject.sntp_test(server)).to eql(false)
         end
 
+        it "returns false if sntp respong no UCST" do
```
teclator: just a typo and maybe reword to `it "returns false if sntp response includes 'no UCST'"`

src/modules/NtpClient.rb
```diff
@@ -788,13 +788,17 @@ def sntp_test(server, ip_version = 4)
...
       # sntp returns always 0 if not called with option -S or -s (set system time)
       # so this is a workaround at least to return false in case server is not
-      # reachable. We could also take care of stdout checking if it includes
-      # "no (U|B)CST reponse", but it also implies be too dependent in the
-      # future and the ntp package should take care of it and aswer other exit
-      # code instead of 0
-      output["stderr"].include?("lookup error") ? false : output["exit"] == 0
+      # reachable.
+      return false if output["stderr"].include?("lookup error")
+      # this check is for valid address, but without ntp. If it break in future
```
teclator: If It breaks in the future

teclator:
LGTM

mvidner:
For the record, I think the commit messages of the form "fixes from code review" are in the same category as "fixes" and "improvements" or "stuff".
Anyway, this instance is harmless enough.

PR #65, opened 2016-11-28
CFA model for ntp conf WIP

joseivanlopez:
CFA model for ntp conf file

### DIFF
--- a/package/yast2-ntp-client.spec
+++ b/package/yast2-ntp-client.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-ntp-client
-Version:        3.2.0
+Version:        3.2.1
 Release:        0
 Summary:        YaST2 - NTP Client Configuration
 License:        GPL-2.0+
--- a/package/yast2-ntp-client.spec
+++ b/package/yast2-ntp-client.spec
@@ -31,6 +31,8 @@ BuildRequires:  yast2-country-data
 BuildRequires:  yast2-devtools >= 3.1.10
 BuildRequires:  rubygem(%rb_default_ruby_abi:rspec)
 BuildRequires:  rubygem(%rb_default_ruby_abi:yast-rake)
+BuildRequires:  augeas-lenses
+BuildRequires:  rubygem(%rb_default_ruby_abi:cfa) >= 0.4.2
 
 #SLPAPI.pm 
 # Hostname::CurrentDomain
@@ -38,6 +40,8 @@ BuildRequires:  rubygem(%rb_default_ruby_abi:yast-rake)
 Requires:       yast2 >= 3.1.11
 Requires:       yast2-country-data
 Requires:       yast2-ruby-bindings >= 1.0.0
+Requires:       rubygem(%rb_default_ruby_abi:cfa) >= 0.4.2
+Requires:       augeas-lenses
 BuildArch:      noarch
 # New sntp command line syntax
 Conflicts:      ntp < 4.2.8
--- a/src/lib/cfa/ntp_conf.rb
+++ b/src/lib/cfa/ntp_conf.rb
@@ -139,6 +139,20 @@ module CFA
 
       attr_reader :tree_data
 
+      def self.parse_augeas(augeas_value)
+        case augeas_value
+        when CFA::AugeasTreeValue
+          value = augeas_value.value
+          tree_data = augeas_value.tree.data
+        else
+          value = augeas_value
+          tree_data = []
+        end
+        [value, tree_data]
+      end
+
+      private_class_method :parse_augeas
+
       private
 
       def add_value(value)
@@ -150,19 +164,7 @@ module CFA
       end
 
       def add_comment(comment)
-        @tree_data << {key: "#comment", value: comment} if !comment.nil?
-      end
-
-      def self.parse_augeas(augeas_value)
-        case augeas_value
-        when CFA::AugeasTreeValue
-          value = augeas_value.value
-          tree_data = augeas_value.tree.data
-        else
-          value = augeas_value
-          tree_data = []
-        end
-        [value, tree_data]
+        @tree_data << { key: "#comment", value: comment } if !comment.nil?
       end
 
       def create_augeas
--- a/src/lib/cfa/ntp_conf.rb
+++ b/src/lib/cfa/ntp_conf.rb
@@ -133,7 +133,7 @@ module CFA
         other.tree_data == self.tree_data
       end
 
-      alias :eq? :==
+      alias_method :eq?, :==
 
       protected
 
--- a/src/lib/cfa/ntp_conf.rb
+++ b/src/lib/cfa/ntp_conf.rb
@@ -6,15 +6,19 @@ module CFA
   class NtpConf < ::CFA::BaseModel
     PARSER = CFA::AugeasParser.new("ntp.lns")
     DEFAULT_PATH = "/etc/ntp.conf".freeze
+    SIMPLE_SETTINGS = ["logfile", "driftfile"].freeze
+    AUTH_COMMANDS = ["trustedkey", "requestkey", "controlkey", "keys"].freeze
     COMMAND_RECORDS = ["server", "peer", "broadcast", "manycastclient", "multicastclient", "manycastserver"].freeze
 
-    attributes(
-      driftfile: "driftfile",
-      logfile: "logfile",
-      keys: "keys",
-      requestkey: "requestkey",
-      controlkey: "controlkey"
-    )
+    def self.ntp_attributes(*attrs)
+      attrs.each do |attr|
+        define_method(attr) {get_attribute(attr.to_s)}
+        define_method("#{attr}=") {|record| set_attribute(attr.to_s, record)}
+        define_method("delete_#{attr}") {delete_attribute(attr.to_s)}
+      end
+    end
+
+    ntp_attributes :driftfile, :logfile, :keys, :requestkey, :controlkey
       
     def initialize(path: DEFAULT_PATH, file_handler: nil)
       super(PARSER, path, file_handler: file_handler)
@@ -32,7 +36,24 @@ module CFA
       RecordCollection.new(RestrictRecord, data.collection("restrict"))
     end
 
+    private
+
+    def get_attribute(name)
+      return nil if data[name].nil? 
+      AttributeRecord.new_from_augeas(data[name])
+    end
+
+    def set_attribute(name, record)
+      data[name] = record.to_augeas
+    end
+
+    def delete_attribute(name)
+      data.delete(name)
+    end
+
     class RecordCollection
+      include Enumerable
+
       def initialize(record_class, augeas_collection) 
         @record_class = record_class
         @augeas_collection = augeas_collection
@@ -40,12 +61,6 @@ module CFA
 
       def each
         @augeas_collection.each do |augeas_value| 
-          yield @record_class.new_from_augeas(augeas_value)
-        end
-      end
-
-      def map
-        @augeas_collection.map do |augeas_value| 
           yield @record_class.new_from_augeas(augeas_value)
         end
       end
@@ -59,62 +74,105 @@ module CFA
       end
 
       def replace(old_record, new_record)
+        return if !self.include? old_record
         matcher = CFA::Matcher.new {|k, v| v == old_record.to_augeas}
         placer = CFA::ReplacePlacer.new(matcher)
         @augeas_collection.add(new_record.to_augeas, placer)
       end
+
+      def empty?
+        @augeas_collection.empty?
+      end
+
+      def ==(other)
+        other.to_a == self.to_a
+      end
+
     end
 
     class Record
-      def self.new_from_augeas(augeas_value)
-        record = self.new
-        record.send(:set_attributes, augeas_value)
-        record
-      end
-
       def initialize(value: nil, tree_data: [])
         @value = value
         @tree_data = tree_data
       end
 
+      def self.new_from_augeas(augeas_value)
+        value, tree_data = self.parse_augeas(augeas_value)
+        record = self.new
+        record.send(:add_value, value)
+        record.send(:add_tree_data, tree_data)
+        record
+      end
+
       attr_reader :value
+
+      def to_augeas
+        create_augeas
+      end
 
       def comment
         comment = @tree_data.select {|d| d[:key] == "#comment"}.first
         comment[:value] if !comment.nil?
       end
 
-      def to_augeas
-        return value if @tree_data.empty?
-        CFA::AugeasTreeValue.new(create_tree, value)
+      def ==(other)
+        other.class == self.class &&
+        other.value == self.value &&
+        other.tree_data == self.tree_data
       end
+
+      alias :eq? :==
+
+      protected
+
+      attr_reader :tree_data
 
       private
 
-      def set_attributes(augeas_value)
-        case augeas_value
-        when CFA::AugeasTreeValue
-          @value = augeas_value.value
-          @tree_data = augeas_value.tree.data
-        else
-          @value = augeas_value
-        end
+      def add_value(value)
+        @value = value
+      end
+
+      def add_tree_data(tree_data)
+        @tree_data = tree_data
       end
 
       def add_comment(comment)
         @tree_data << {key: "#comment", value: comment} if !comment.nil?
       end
 
-      def create_tree
+      def self.parse_augeas(augeas_value)
+        case augeas_value
+        when CFA::AugeasTreeValue
+          value = augeas_value.value
+          tree_data = augeas_value.tree.data
+        else
+          value = augeas_value
+          tree_data = []
+        end
+        [value, tree_data]
+      end
+
+      def create_augeas
+        return value if @tree_data.empty?
         tree = CFA::AugeasTree.new
         @tree_data.each {|d| tree.add(d[:key], d[:value])}
-        tree
+        CFA::AugeasTreeValue.new(tree, value)
+      end
+    end
+
+    class AttributeRecord < Record
+      def initialize(value: nil, comment: nil)
+        super()
+        add_value(value)
+        add_comment(comment)
       end
     end
 
     class CommandRecord < Record
       def initialize(value: nil, options: [], comment: nil)
-        super(value: value)
+        super()
+        add_value(value)
         add_options(options)
         add_comment(comment)
       end
@@ -139,7 +197,8 @@ module CFA
 
     class RestrictRecord < Record
       def initialize(value: nil, actions: [], comment: nil)
-        super(value: value)
+        super()
+        add_value(value)
         add_actions(actions)
         add_comment(comment)
       end
--- a/src/lib/cfa/ntp_conf.rb
+++ b/src/lib/cfa/ntp_conf.rb
@@ -3,6 +3,10 @@ require "cfa/augeas_parser"
 require "cfa/matcher"
 
 module CFA
+  # class representings /etc/ntp.conf file model. It provides helper to manipulate
+  # with file. It uses CFA framework and Augeas parser.
+  # @see http://www.rubydoc.info/github/config-files-api/config_files_api/CFA/BaseModel
+  # @see http://www.rubydoc.info/github/config-files-api/config_files_api/CFA/AugeasParser
   class NtpConf < ::CFA::BaseModel
     PARSER = CFA::AugeasParser.new("ntp.lns")
     DEFAULT_PATH = "/etc/ntp.conf".freeze
@@ -51,6 +55,13 @@ module CFA
       data.delete(name)
     end
 
+    # class to manage a collection of some ntp entries, as
+    # server, restrict, peer, etc.
+    #   For example:
+    #     server 0.opensuse.pool.ntp.org iburst
+    #     server 1.opensuse.pool.ntp.org iburst
+    #     server 2.opensuse.pool.ntp.org iburst
+    #     server 3.opensuse.pool.ntp.org iburst
     class RecordCollection
       include Enumerable
 
@@ -90,6 +101,7 @@ module CFA
 
     end
 
+    # class to represent a general entry of the file.
     class Record
       def initialize(value: nil, tree_data: [])
         @value = value
@@ -161,6 +173,9 @@ module CFA
       end
     end
 
+    # class to represent a key-value ntp entry.
+    #   For example:
+    #     logfile /var/log/ntp
     class AttributeRecord < Record
       def initialize(value: nil, comment: nil)
         super()
@@ -169,6 +184,10 @@ module CFA
       end
     end
 
+    # class to represent ntp command entries, as
+    # server, peer.
+    #   For example:
+    #     server 0.opensuse.pool.ntp.org iburst
     class CommandRecord < Record
       def initialize(value: nil, options: [], comment: nil)
         super()
@@ -195,6 +214,9 @@ module CFA
     class PeerRecord < CommandRecord
     end
 
+    # class to represent ntp restrict entries.
+    #   For example:
+    #     restrict -4 default notrap nomodify nopeer noquery
     class RestrictRecord < Record
       def initialize(value: nil, actions: [], comment: nil)
         super()
--- a/src/lib/cfa/ntp_conf.rb
+++ b/src/lib/cfa/ntp_conf.rb
@@ -7,6 +7,24 @@ module CFA
   # with file. It uses CFA framework and Augeas parser.
   # @see http://www.rubydoc.info/github/config-files-api/config_files_api/CFA/BaseModel
   # @see http://www.rubydoc.info/github/config-files-api/config_files_api/CFA/AugeasParser
+  #
+  # In NTP files some keys can be present multiple times.
+  #
+  # For example:
+  #
+  # server 1.pool.ntp.org iburst
+  # server 2.pool.ntp.org
+  # server 127.127.1.0
+  # fudge 127.127.1.0 stratum 10
+  # peer 128.100.0.45
+  # peer 192.168.1.30
+  #
+  # Order of lines is not important, except for the
+  # fudge option that should immediately follow
+  # appropriate "server" option.
+  #
+  # Important keys (that are to be read and written) are
+  # server, peer, fudge, broadcast and broadcastclient.
   class NtpConf < ::CFA::BaseModel
     PARSER = CFA::AugeasParser.new("ntp.lns")
 
@@ -29,15 +47,43 @@ module CFA
       super(PARSER, PATH, file_handler: file_handler)
     end
 
+    # Loads and fixes AugeasElement keys.
+    #
+    # AugeasParser returns collection entries with
+    # key ending in '[]', for example 'server[]'.
+    #
+    # If only exists one entry of the collection, parser
+    # return key without '[]'.
+    #
+    # For example:
+    #
+    # server 1.pool.ntp.org iburst
+    # server 2.pool.ntp.org
+    # peer 128.100.0.45
+    #
+    # In this case, key for peer entry is 'peer'.
+    # It is necessary to add '[]' to avoid possible
+    # write issues when more entries of that type
+    # are added.
     def load
       super
       fix_keys
     end
 
+    # Obtains a collection that represents the
+    # entries of the file.
+    # @return [CollectionRecord] collection to
+    #   manipulate ntp entries.
+    #
+    # The collection preserves the order of the
+    # ntp entries in the file and only contains the
+    # entries of interest (see RECORD_ENTRIES).
     def records
       RecordCollection.new(data)
     end
 
+    # Obtains raw content of the file.
+    # @return [String]
     def raw
       PARSER.serialize(data)
     end
@@ -61,7 +107,10 @@ module CFA
       end
     end
 
-    # class to manage ntp entries
+    # class to manage ntp entries as a collection.
+    #
+    # Each element of the collection is a subclass of Record.
+    # A Record object is a wrapper of an AugeasElement.
     class RecordCollection
       include Enumerable
 
@@ -75,12 +124,16 @@ module CFA
         end
       end
 
+      # Adds a new Record object to the collection.
+      # @param [Record] record
       def add(record)
         @augeas_tree.data << record.augeas
       end
 
       alias_method :<<, :add
 
+      # Removes a Record object from the collection.
+      # @param [Record] record
       def delete(record)
         matcher = Matcher.new do |k, v|
           k == record.augeas[:key] &&
@@ -113,17 +166,30 @@ module CFA
     end
 
     # Base class to represent a general ntp entry.
+    #
+    # This class is an AugeasElement wrapper. A Record
+    # has a value, and could also contains a comment and
+    # options. Its contain is stored in an AugeasElement.
+    #
+    # Each ntp entry type has different interpretation
+    # for its options in the AugeasElement. For each one,
+    # a Record subclass is created.
     class Record
-      def self.record_class(key)
-        entry_type = key.gsub("[]", "")
-        record_class = ["::CFA::NtpConf::", entry_type.capitalize, "Record"].join
-        Kernel.const_get(record_class)
-      end
-
+      # Creates the corresponding subclass object according
+      # to its AugeasElement key.
+      # @param [String] key
       def self.new_from_augeas(augeas_entry)
         record = record_class(augeas_entry[:key]).new
         record.augeas = augeas_entry
         record
+      end
+
+      # Returns the corresponding subclass
+      # @param [string] key
+      def self.record_class(key)
+        entry_type = key.gsub("[]", "")
+        record_class = ["::CFA::NtpConf::", entry_type.capitalize, "Record"].join
+        Kernel.const_get(record_class)
       end
 
       def initialize(key = nil)
@@ -169,10 +235,12 @@ module CFA
         augeas[:key].gsub("[]", "")
       end
 
+      # Returns an String representing the options
       def raw_options
         options.join(" ")
       end
 
+      # Sets options from a String
       def raw_options=(raw_options)
         self.options = split_raw_options(raw_options)
       end
@@ -204,8 +272,8 @@ module CFA
       end
     end
 
-    # Base class to represent a ntp command entry, as
-    # server, peer, broadcast, etc.
+    # class to represent a ntp command record entry. There is a
+    # subclass for server, peer, broadcast and broacastclient.
     class CommandRecord < Record
       def options
         return [] unless tree_value?
@@ -220,8 +288,8 @@ module CFA
     end
 
     # class to represent a ntp server entry.
-    #   For example:
-    #     server 0.opensuse.pool.ntp.org iburst
+    # For example:
+    #   server 0.opensuse.pool.ntp.org iburst
     class ServerRecord < CommandRecord
       def initialize
         super("server[]")
@@ -229,8 +297,8 @@ module CFA
     end
 
     # class to represent a ntp peer entry.
-    #   For example:
-    #     peer 128.100.0.45
+    # For example:
+    #   peer 128.100.0.45
     class PeerRecord < CommandRecord
       def initialize
         super("peer[]")
@@ -238,8 +306,8 @@ module CFA
     end
 
     # class to represent a ntp broadcast entry.
-    #   For example:
-    #     broadcast 128.100.0.45
+    # For example:
+    #   broadcast 128.100.0.45
     class BroadcastRecord < CommandRecord
       def initialize
         super("broadcast[]")
@@ -254,8 +322,11 @@ module CFA
     end
 
     # class to represent a ntp fudge entry.
-    #   For example:
-    #     fudge  127.127.1.0 stratum 10
+    #
+    # For example:
+    #   fudge  127.127.1.0 stratum 10
+    #
+    # Fudge entry has its own options interpretation.
     class FudgeRecord < CommandRecord
       def initialize
         super("fudge[]")
@@ -285,8 +356,11 @@ module CFA
     end
 
     # class to represent a ntp restrict entry.
-    #   For example:
-    #     restrict -4 default notrap nomodify nopeer noquery
+    #
+    # For example:
+    #   restrict -4 default notrap nomodify nopeer noquery
+    #
+    # Restrict entry has its own options interpretation.
     class RestrictRecord < Record
       def initialize
         super("restrict[]")
--- a/src/lib/cfa/ntp_conf.rb
+++ b/src/lib/cfa/ntp_conf.rb
@@ -88,7 +88,7 @@ module CFA
       PARSER.serialize(data)
     end
 
-    private
+  private
 
     def fix_keys
       fix_collection_keys(data.data)
@@ -155,7 +155,7 @@ module CFA
         other.to_a == to_a
       end
 
-      private
+    private
 
       def record_entries
         matcher = Matcher.new do |k, _v|
@@ -245,7 +245,7 @@ module CFA
         self.options = split_raw_options(raw_options)
       end
 
-      private
+    private
 
       def tree_value?
         @augeas[:value].is_a?(AugeasTreeValue)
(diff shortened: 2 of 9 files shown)


src/lib/cfa/ntp_conf.rb
```diff
@@ -0,0 +1,248 @@
...
+      attrs.each do |attr|
+        define_method(attr) { get_attribute(attr.to_s) }
+        define_method("#{attr}=") { |record| set_attribute(attr.to_s, record) }
+        define_method("delete_#{attr}") { delete_attribute(attr.to_s) }
+      end
+    end
+
+    ntp_attributes :driftfile, :logfile, :keys, :requestkey, :controlkey
+
+    def initialize(path: DEFAULT_PATH, file_handler: nil)
```
jreidinger: do you see usecase where overwrite of path is needed?

joseivanlopez: Not really.

src/lib/cfa/ntp_conf.rb
```diff
@@ -0,0 +1,248 @@
...
+      AttributeRecord.new_from_augeas(data[name])
+    end
+
+    def set_attribute(name, record)
+      data[name] = record.to_augeas
+    end
+
+    def delete_attribute(name)
+      data.delete(name)
+    end
```
jreidinger: well, this smells a bit from code POV. ruby do not use get/set and also all have attribute as common, which indicate that it will probably be better if it is method of attribute object. 
also whole this handling is a bit duplicate to existing API https://github.com/config-files-api/config_files_api/blob/master/lib/cfa/base_model.rb#L97

so, if you need more complex functionality then simple get value/ set value I suggest to have attribute as object like

```
class Attribute # probably merge with AttributeRecord class
  def initialize(name, parent)
    @name = name
    @data = parent
  end

  def exists?
    !!@data[@name]
  end

  def value=(value)
    @data[@name] = value.to_augeast
  end

  ...
end
```

src/lib/cfa/ntp_conf.rb
```diff
@@ -0,0 +1,248 @@
...
+        @augeas_collection.add(record.to_augeas)
+      end
+
+      def delete(record)
+        @augeas_collection.delete(record.to_augeas)
+      end
+
+      def replace(old_record, new_record)
+        return if !include? old_record
+        matcher = CFA::Matcher.new { |_k, v| v == old_record.to_augeas }
```
jreidinger: wrong matcher, I think better to use `matcher = CFA::Matcher.new(value_matcher: old_record.to_augeas)`

src/lib/cfa/ntp_conf.rb
```diff
@@ -0,0 +1,248 @@
...
+      def add(record)
+        @augeas_collection.add(record.to_augeas)
+      end
+
+      def delete(record)
+        @augeas_collection.delete(record.to_augeas)
+      end
+
+      def replace(old_record, new_record)
+        return if !include? old_record
```
jreidinger: what about using `unless include?(old_record)`

src/lib/cfa/ntp_conf.rb
```diff
@@ -0,0 +1,248 @@
...
+      end
+
+      def empty?
+        @augeas_collection.empty?
+      end
+
+      def ==(other)
+        other.to_a == to_a
+      end
+    end
```
jreidinger: In the end to have it mergable it really need documentation with all constraints on arguments and also return value.

src/lib/cfa/ntp_conf.rb
```diff
@@ -0,0 +1,248 @@
...
+    # class to represent a general entry of the file.
+    class Record
+      def initialize(value: nil, tree_data: [])
+        @value = value
+        @tree_data = tree_data
+      end
+
+      def self.new_from_augeas(augeas_value)
+        value, tree_data = parse_augeas(augeas_value)
+        record = new
```
jreidinger: why not `new(value: value, tree_data: tree_data)`? I really do not like ignoring private accessor.

src/lib/cfa/ntp_conf.rb
```diff
@@ -0,0 +1,248 @@
...
+
+      attr_reader :value
+
+      def to_augeas
+        create_augeas
+      end
+
+      def comment
+        comment = @tree_data.select { |d| d[:key] == "#comment" }.first
+        comment[:value] if !comment.nil?
```
jreidinger: what if there is multiple comments? it looks wrong for me

src/lib/cfa/ntp_conf.rb
```diff
@@ -0,0 +1,248 @@
...
+      end
+
+      def ==(other)
+        other.class == self.class &&
+          other.value == value &&
+          other.tree_data == tree_data
+      end
+
+      alias_method :eq?, :==
```
jreidinger: eq or eql?

src/lib/cfa/ntp_conf.rb
```diff
@@ -0,0 +1,248 @@
...
+
+      private
+
+      def add_value(value)
+        @value = value
+      end
+
+      def add_tree_data(tree_data)
+        @tree_data = tree_data
+      end
```
jreidinger: what is goal of this two private methods? and it looks like attr_writers.

src/lib/cfa/ntp_conf.rb
```diff
@@ -0,0 +1,248 @@
...
+
+      def add_comment(comment)
+        @tree_data << { key: "#comment", value: comment } if !comment.nil?
+      end
+
+      def create_augeas
```
jreidinger: where it is used?

src/lib/cfa/ntp_conf.rb
```diff
@@ -0,0 +1,248 @@
...
+    end
+
+    # class to represent a key-value ntp entry.
+    #   For example:
+    #     logfile /var/log/ntp
+    class AttributeRecord < Record
+      def initialize(value: nil, comment: nil)
+        super()
+        add_value(value)
+        add_comment(comment)
```
jreidinger: add_comment is private, not protected. Also it looks a bit strange to have child that have so different arguments in constructor

jreidinger:
well, in general there is a lot of new API which looks quite good, but I would like to see at least one or two real usage in code, so I can see its pros&cons

test/cfa/ntp_conf_test.rb
```diff
@@ -0,0 +1,365 @@
...
+  subject(:ntp) { described_class.new(file_handler: ntp_file) }
+
+  let(:ntp_file) { CFA::MemoryFile.new(content) }
+
+  before do
+    ntp.load
+  end
+
+  context "when ask for the collection" do
+    context "#servers" do
```
imobachgs: Although [#describe](https://github.com/rspec/rspec-core/blob/master/lib/rspec/core/example_group.rb#L272) and [#context](https://github.com/rspec/rspec-core/blob/master/lib/rspec/core/example_group.rb#L277) just aliases, I rather prefer using `describe` in this case. You're not setting the context but describing the method.

src/lib/cfa/ntp_conf.rb
```diff
@@ -0,0 +1,387 @@
...
+    # The collection preserves the order of the
+    # ntp entries in the file and only contains the
+    # entries of interest (see RECORD_ENTRIES).
+    def records
+      RecordCollection.new(data)
+    end
+
+    # Obtains raw content of the file.
+    # @return [String]
+    def raw
```
jreidinger: why this method is needed?

src/lib/cfa/ntp_conf.rb
```diff
@@ -0,0 +1,387 @@
...
+        if entry[:value].is_a?(AugeasTreeValue)
+          fix_collection_keys(entry[:value].tree.data)
+        end
+      end
+    end
+
+    def fix_collection_keys(entries)
+      entries.each do |entry|
+        if COLLECTION_KEYS.include?(entry[:key])
+          entry[:key] += "[]" unless entry[:key] =~ /.*\[\]$/
```
jreidinger: it is better to prefer `<<` over `+=` as it is more efficient ( new copy is not needed ), especially if you do mass processing.
And also I prefer `entry[:key].end_with?("[]")` over `entry[:key] =~ /.*\[\]$/` as it is more readable and efficient.

jreidinger: and BTW this never can happen as COLLECTION_KEYS do not contain names with `[]`

src/lib/cfa/ntp_conf.rb
```diff
@@ -0,0 +1,387 @@
...
+    end
+
+  private
+
+    def fix_keys
+      fix_collection_keys(data.data)
+      data.data.each do |entry|
+        if entry[:value].is_a?(AugeasTreeValue)
+          fix_collection_keys(entry[:value].tree.data)
+        end
```
jreidinger: in fact I do not get, why two methods are needed. I expect one method with something like
```
def fix_keys(data)
  data.each do |entry|
    entry[:key] += "[]" if COLLECTION_KEYS.include?(entry[:key])
    fix_keys(entry[:value].tree.data) if entry[:value].is_a?(AugeasTreeValue)
  end
end
```

src/lib/cfa/ntp_conf.rb
```diff
@@ -0,0 +1,387 @@
...
+      end
+
+      # Adds a new Record object to the collection.
+      # @param [Record] record
+      def add(record)
+        @augeas_tree.data << record.augeas
+      end
+
+      alias_method :<<, :add
```
jreidinger: what is reason to have two same methods? I do not see in Array ( as this class acts like array ) method add. So I prefer to have only `<<` and if you really need additional method, then better one is `push` but it allows multiple arguments.

src/lib/cfa/ntp_conf.rb
```diff
@@ -0,0 +1,387 @@
...
+
+      # Removes a Record object from the collection.
+      # @param [Record] record
+      def delete(record)
+        matcher = Matcher.new do |k, v|
+          k == record.augeas[:key] &&
+            v == record.augeas[:value]
+        end
```
jreidinger: `matcher = Matcher.new(key: record.augeas[:key], value_matcher: record.augeas[:value])`

src/lib/cfa/ntp_conf.rb
```diff
@@ -0,0 +1,387 @@
...
+    class RecordCollection
+      include Enumerable
+
+      def initialize(augeas_tree)
+        @augeas_tree = augeas_tree
+      end
+
+      def each
+        record_entries.each do |augeas_element|
+          yield Record.new_from_augeas(augeas_element)
```
jreidinger: hmm, this means that any method from Enumerable creates bunch of new object. It will very stress GC and also in general ineffecient. As here only available manipulators are `<<,delete, delete_if` I suggest to construct once and recreate only when manipulated. So something like

```
def each(&block)
  record_entries.each(&block)
end

def delete
  ...
  reset_cache
end

def reset_cache
  @record_entries = nil
end

def record_entries
   return @record_entries if @record_entries
  matcher = Matcher.new do |k, _v|
    RECORD_ENTRIES.include?(k.gsub("[]", ""))
  end
  @record_entries = @augeas_tree.select(matcher).map { |e| Record.new_from_augeas(augeas_element) }
end
```
I hope you get idea from this snippet.

src/lib/cfa/ntp_conf.rb
```diff
@@ -0,0 +1,387 @@
...
+        @augeas_tree.delete(matcher)
+      end
+
+      def delete_if(&block)
+        records = select(&block)
+        records.each { |record| delete(record) }
+      end
+
+      def empty?
+        count == 0
```
jreidinger: ah, that also kills performance. if you go with cache as I suggest, then it is simple about delegator to pass methods to `record_entries`. just to get idea, each call to empty? in your case call count, which in enurable call yield with block that just do +1. So to recognize if recordCollection is empty now you have to create X new object, where X is number of records in config file, which is killer for average sized files.

src/lib/cfa/ntp_conf.rb
```diff
@@ -0,0 +1,387 @@
...
+      end
+
+      def ==(other)
+        other.to_a == to_a
```
jreidinger: well, above you mention that order is not important, but here is order dependent comparison, is it ok?

src/lib/cfa/ntp_conf.rb
```diff
@@ -0,0 +1,387 @@
...
+        record.augeas = augeas_entry
+        record
+      end
+
+      # Returns the corresponding subclass
+      # @param [string] key
+      def self.record_class(key)
+        entry_type = key.gsub("[]", "")
+        record_class = ["::CFA::NtpConf::", entry_type.capitalize, "Record"].join
+        Kernel.const_get(record_class)
```
jreidinger: hmm, quite magic, but I think we can do it at least without join like `record_class = "::CFA::NtpConf::#{entry_type.capitalize}Record"

src/lib/cfa/ntp_conf.rb
joseivanlopez: Because in log is printed the content of the file

src/lib/cfa/ntp_conf.rb
```diff
@@ -0,0 +1,387 @@
...
+      end
+
+      def initialize(key = nil)
+        @augeas = { key: key, value: nil }
+      end
+
+      attr_accessor :augeas
```
jreidinger: is it good idea to have accessor over whole `@augeas` ? I think it is really easy to make class very inconsistent. Any change to `:key` can be very critical and any change to `:value` can kill whole subtree concept introduced by other methods.
What I suggest is to have `key, value and value=` methods.

src/lib/cfa/ntp_conf.rb
```diff
@@ -0,0 +1,387 @@
...
+        end
+        @augeas_tree.select(matcher)
+      end
+    end
+
+    # Base class to represent a general ntp entry.
+    #
+    # This class is an AugeasElement wrapper. A Record
+    # has a value, and could also contains a comment and
+    # options. Its contain is stored in an AugeasElement.
```
jreidinger: what do you mean with `Its contain is stored in an AugeasElement.` ?
What I see in general as problem is that if you do not work on augeas tree, then it can quite easily get into very inconsistent state where record contain some changes, but it is never propagated. What I suggest is to directly operate on augeas tree.

For example see how BooleanAttribute operate in cfa itself https://github.com/config-files-api/config_files_api/blob/master/lib/cfa/base_model.rb#L157

to demonstrate it try 
```
collection.to_a[1].value = "test"
collection << Record.new("server", "192.168.0.1")
puts collection.to_a[1].value
```
I hope you see what I mean.

src/lib/cfa/ntp_conf.rb
```diff
@@ -0,0 +1,387 @@
...
+      end
+
+    private
+
+      def tree_value?
+        @augeas[:value].is_a?(AugeasTreeValue)
+      end
+
+      def tree_value
+        tree_value? ? @augeas[:value] : nil
```
jreidinger: what is goal of this check? later and also previously method is called on result, so in fact it will call `nil.tree` or `nil.value` which will fail. If you want check then raise exception or ignore check and simply return `@augeas[:value]`

src/lib/cfa/ntp_conf.rb
```diff
@@ -0,0 +1,387 @@
...
+        Matcher.new { |k, _v| !k.include?("#comment") }
+      end
+    end
+
+    # class to represent a ntp command record entry. There is a
+    # subclass for server, peer, broadcast and broacastclient.
+    class CommandRecord < Record
+      def options
+        return [] unless tree_value?
+        augeas_options.map { |option| option[:key] }
```
jreidinger: hmm, you call private method here. It should not work, but it looks like there is a bug in ruby itself, which I hope they fix later.

my testing code which proves bug
```
class C
private
  def a
    puts "a"
  end
end

class D < C
  def b
    a
  end
end

D.new.b
```

src/lib/cfa/ntp_conf.rb
```diff
@@ -0,0 +1,387 @@
...
+      end
+
+      def options=(options)
+        create_tree_value unless tree_value?
```
jreidinger: as I mentioned above if you have `ensure_tree_value` then these conditions need not to be repeated.

src/lib/cfa/ntp_conf.rb
```diff
@@ -0,0 +1,387 @@
...
+      end
+
+      # Returns an String representing the options
+      def raw_options
+        options.join(" ")
+      end
+
+      # Sets options from a String
+      def raw_options=(raw_options)
+        self.options = split_raw_options(raw_options)
```
jreidinger: where is this options= defined? it looks for me like it is in child, but not in everyone, so actually this method ends in strange error for some sub-classes which do not look nice.

src/modules/NtpClient.rb
```diff
@@ -282,6 +296,38 @@ def GetNtpServersByCountry(country, terse_output)
...
+      if !File.exist?(NTP_FILE)
+        log.error("File #{NTP_FILE} does not exist")
+        return false
+      end
+
+      begin
+        @ntp_conf = CFA::NtpConf.new
+        @ntp_conf.load
+      rescue
+        log.error("Failed to read #{NTP_FILE}")
```
jreidinger: if you log message, it should be useful. I think at least write error message here to log would be very helpful as it can be bad permissions, corrupted disk, malformed file or whatever else.

test/cfa/ntp_conf_test.rb
```diff
@@ -0,0 +1,305 @@
...
+
+  let(:ntp_file) { CFA::MemoryFile.new(ntp_disk_file) }
+
+  let(:new_record) do
+    record = CFA::NtpConf::ServerRecord.new
+    record.value = "4.pool.ntp.org"
+    record
+  end
+
+  let(:existing_record) { ntp.records.first.dup }
```
jreidinger: why dup is needed?

test/cfa/ntp_conf_test.rb
```diff
@@ -0,0 +1,305 @@
...
+
+  before do
+    ntp.load
+  end
+
+  context "#add" do
+    context "when does not exist the record to add" do
+      it "adds the record" do
+        ntp.records.add(new_record)
+        expect(ntp.records.include?(new_record)).to be(true)
```
jreidinger: to be consistent I think it would be nice if it is there also count to ensure it is there only once.

test/cfa/ntp_conf_test.rb
```diff
@@ -0,0 +1,305 @@
...
+    end
+  end
+
+  describe "#value=" do
+    it "sets the value of the record" do
+      value = "1.pool.ntp.org"
+      record.value = value
+      expect(record.value).to eq(value)
+    end
+  end
```
jreidinger: and here I am curious how it is propagated back to augeas tree?

test/cfa/ntp_conf_test.rb
```diff
@@ -0,0 +1,305 @@
...
+
+  subject(:record) { described_class.new_from_augeas(augeas_element) }
+
+  describe "#options" do
+    it "obtains the options of the record" do
+      expect(record.options).to eq(["iburst"])
+    end
+  end
+
+  describe "#options=" do
```
jreidinger: and vice verse for setting options. I would like to see that new options are saved and then verify that it looks as expected in ntp.conf. ( actually that is purpose of memoryFile file handler, for easier testing read/write )

src/lib/cfa/ntp_conf.rb
jreidinger: ok, makes sense. I am just not sure if raw is good name, but do not have in mind better.

src/lib/cfa/ntp_conf.rb
joseivanlopez: Yes. We are considering two "files" identical if they have the same lines in the same order.

src/lib/cfa/ntp_conf.rb
joseivanlopez: Yes, better.

src/lib/cfa/ntp_conf.rb
joseivanlopez: ok

src/lib/cfa/ntp_conf.rb
joseivanlopez: Really, Record class is working on augeas tree, and this is what I didn't explain correctly. I think that the examaple you propose it is correct.

src/lib/cfa/ntp_conf.rb
joseivanlopez: I think it is correct in ruby. In ruby, access control is slighly different to other languages as C++. In fact, private means it only can be passed to self.

src/lib/cfa/ntp_conf.rb
```diff
@@ -0,0 +1,387 @@
...
+        end
+      end
+
+      def comment
+        return nil unless tree_value?
+        tree_value.tree["#comment"]
+      end
+
+      def comment=(comment)
+        create_tree_value unless tree_value?
```
joseivanlopez: Ok

src/lib/cfa/ntp_conf.rb
joseivanlopez: All Record subclasses have to define this method. Perhaps, Record also has to define the method with a raise message indicating that subclases have to define it.

test/cfa/ntp_conf_test.rb
joseivanlopez: Done

test/cfa/ntp_conf_test.rb
joseivanlopez: A Record is a wrapper of an AugeasElement contained into the AugeasTree of the file. When we change the record, its AugeasElement is modified. For this reason, the changes are always propagated back.

test/cfa/ntp_conf_test.rb
```diff
@@ -0,0 +1,305 @@
...
+  describe "#comment" do
+    context "when the record has not comment" do
+      it "obtains nil" do
+        expect(record.comment).to eq(nil)
+      end
+    end
+
+    context "when the record has comment" do
+      let(:augeas_options) do
+        tree = CFA::AugeasTree.new
```
joseivanlopez: Ok

test/cfa/ntp_conf_test.rb
joseivanlopez: Done

src/lib/cfa/ntp_conf.rb
joseivanlopez: Done

src/lib/cfa/ntp_conf.rb
joseivanlopez: Done

src/lib/cfa/ntp_conf.rb
```diff
@@ -0,0 +1,387 @@
...
+        matcher = Matcher.new do |k, v|
+          k == record.augeas[:key] &&
+            v == record.augeas[:value]
+        end
+        @augeas_tree.delete(matcher)
+      end
+
+      def delete_if(&block)
+        records = select(&block)
+        records.each { |record| delete(record) }
```
joseivanlopez: I added this method because a user of the ntp model should not use an augeas matcher directly, I supose.

src/lib/cfa/ntp_conf.rb
joseivanlopez: Ok

src/lib/cfa/ntp_conf.rb
joseivanlopez: Ok. I remove add

src/modules/NtpClient.rb
joseivanlopez: Ok. I will print the exception message.

src/lib/cfa/ntp_conf.rb
jreidinger: I am fine with this method, it makes sense. I just do not like its implementation as inefficient.

src/lib/cfa/ntp_conf.rb
jreidinger: so then what is purpose of `protected` if it act like private? 
it is a bit explained in https://tenderlovemaking.com/2012/09/07/protected-methods-and-ruby-2-0.html so it looks like protected is more like `friend` method for same class. I think in general not good naming in ruby for anyone coming from any different language.

src/lib/cfa/ntp_conf.rb
```diff
@@ -90,20 +90,12 @@ def raw
...
-      end
-    end
-
-    def fix_collection_keys(entries)
-      entries.each do |entry|
+    def fix_keys(tree)
+      tree.data.each do |entry|
         if COLLECTION_KEYS.include?(entry[:key])
-          entry[:key] += "[]" unless entry[:key] =~ /.*\[\]$/
+          entry[:key] << "[]" unless entry[:key].end_with?("[]")
```
jreidinger: as I said this `entry[:key].end_with?("[]")` cannot happen as `COLLECTION_KEYS` does not have `[]` inside, so that include won't match.

src/lib/cfa/ntp_conf.rb
```diff
@@ -119,27 +111,27 @@ def initialize(augeas_tree)
       end
 
       def each
-        record_entries.each do |augeas_element|
-          yield Record.new_from_augeas(augeas_element)
+        record_entries.each do |record|
+          yield record
         end
```
jreidinger: well, problem I have with this implementation is that it kills each ability to act like iterator ( I can give you some articles on this topic )

to demonstrate it try this example code and its output:

```
class C
  def initialize
    @arr = [1,2,3]
  end

  def each(&block)
    @arr.each(&block)
  end
end

o = C.new
o.each { |i| puts i.to_s }
puts o.each.count

class D
  def initialize
    @arr = [1,2,3]
  end

  def each(&block)
    @arr.each do |record|
      yield record
    end
  end
end

o = D.new
o.each { |i| puts i.to_s }
puts o.each.count
```

src/lib/cfa/ntp_conf.rb
```diff
@@ -179,24 +179,23 @@ class Record
...
-        record_class = ["::CFA::NtpConf::", entry_type.capitalize, "Record"].join
+        record_class = "::CFA::NtpConf::#{entry_type.capitalize}Record"
         Kernel.const_get(record_class)
       end
 
-      def initialize(key = nil)
-        @augeas = { key: key, value: nil }
+      def initialize(augeas = nil)
+        augeas ||= create_augeas
+        @augeas = augeas
```
jreidinger: good, now I see how it can act on augeas tree as you pass now reference to it. Previously it was well hidden.

src/lib/cfa/ntp_conf.rb
```diff
@@ -179,24 +179,23 @@ class Record
...
       end
 
-      attr_accessor :augeas
+      attr_reader :augeas
```
jreidinger: I worry it still do not help, as this means, that you can get this reference and modify it on its own, which is from my POV violation of encapsulation. Why it is needed?

src/lib/cfa/ntp_conf.rb
```diff
@@ -212,27 +211,38 @@ def value=(value)
...
 
       def comment=(comment)
-        create_tree_value unless tree_value?
+        ensure_tree_value
         if comment.to_s == ""
-          tree_value.tree.delete("#comment")
+          tree_value.tree.delete("#comment[]")
         else
-          tree_value.tree["#comment"] = comment
+          tree_value.tree["#comment[]"] = comment
```
jreidinger: if there is more comments which one will be overwritten? I think this part need test coverage of this functionality. So we are sure how it exactly works and can document it.

src/lib/cfa/ntp_conf.rb
```diff
@@ -328,8 +346,8 @@ def initialize
     #
     # Fudge entry has its own options interpretation.
     class FudgeRecord < CommandRecord
-      def initialize
-        super("fudge[]")
+      def create_augeas
+        { key: "fudge[]", value: nil }
```
jreidinger: well, when I see this method overwritten it almost always look same and from 
```
def create_augeas
  { key: "fudge[]", value: nil }
end
```

only information holder it "create_augeas" so we know what is doing and "fudge[]" which doesn't look like enough for me and too much non-interesting stuff there. And also it is not instance specific, but class specific.

So what about using something like 
```
class FudgeRecord
  AUGEAS_KEY = "fudge[]"
  ...
end

# and in Record
def create_augeas
  { key: self.class.const_get("AUGEAS_KEY"), value: nil }
end
```

another option is class attribute or class method. I am fine with any of it which simplify kids of Record to hold only relevant information.

test/cfa/ntp_conf_test.rb
```diff
@@ -39,36 +47,80 @@ def ntp_disk_file
...
+    context "when a record is updated" do
+      it "modifies an entry" do
+        record = ntp.records.find { |r| r.value == "0.pool.ntp.org" }
+        record.value = "10.pool.ntp.org"
+        ntp.save
+        expect(file.content.lines).not_to include("server 0.pool.ntp.org\n")
+        expect(file.content.lines).to include("server 10.pool.ntp.org\n")
+      end
+    end
+  end
```
jreidinger: :+1:

test/cfa/ntp_conf_test.rb
```diff
@@ -131,88 +187,98 @@ def ntp_disk_file
...
 
   describe "#comment=" do
     it "sets a comment to the record" do
-      comment = "sample comment"
+      comment = "#sample comment"
+      record = ntp.records.first
       record.comment = comment
       expect(record.comment).to eq(comment)
+      ntp.save
+      expect(file.content).to include("server 0.pool.ntp.org iburst#sample comment\n")
```
jreidinger: is it valid to have no space after iburst? can ntp handle such ntp.conf?

test/cfa/ntp_conf_test.rb
```diff
@@ -131,88 +187,98 @@ def ntp_disk_file
...
     end
   end
 
   describe "#raw_options=" do
     it "sets options from a string" do
+      record = ntp.records.first
       record.raw_options = "iburst prefer"
       expect(record.options).to eq(["iburst", "prefer"])
+      ntp.save
+      expect(file.content).to include("server 0.pool.ntp.org iburst prefer\n")
```
jreidinger: :+1: now it is much more readable and verificable that we really write what we want.

test/cfa/ntp_conf_test.rb
```diff
@@ -287,19 +369,23 @@ def ntp_disk_file
...
+      expect(record.raw_options).to eq("default notrap nomodify nopeer")
     end
   end
 
   describe "#raw_options=" do
     it "sets options from a string" do
       record.raw_options = "default notrap"
       expect(record.options).to eq(["default", "notrap"])
+      ntp.save
+      expect(file.content).to include("restrict -4 default notrap\n")
```
jreidinger: this brings me question, how to change restriction from ipv4 to ipv6? is it possible?

src/modules/NtpClient.rb
```diff
@@ -305,8 +305,8 @@ def read_ntp_conf
       begin
         @ntp_conf = CFA::NtpConf.new
         @ntp_conf.load
-      rescue
-        log.error("Failed to read #{NTP_FILE}")
+      rescue StandardError => e
+        log.error("Failed to read #{NTP_FILE}: #{e.message}")
```
jreidinger: :+1:

src/lib/cfa/ntp_conf.rb
joseivanlopez: It is needed to add a new record to the collection. The idea is to access to Record#augeas for adding it to the AugeasTree, so any change in the record is done in the AugeasTree.

src/lib/cfa/ntp_conf.rb
joseivanlopez: Really, a record only has a comment (line comment), so it should not be a collection. I correct it.

src/lib/cfa/ntp_conf.rb
joseivanlopez: Ok, I agree.

test/cfa/ntp_conf_test.rb
joseivanlopez: Yes, it is valid. In fact, augeas saves it in this form.

test/cfa/ntp_conf_test.rb
```diff
@@ -131,88 +187,98 @@ def ntp_disk_file
...
+        record = ntp.records.to_a[1]
+        expect(record.comment).to eq("#sample comment")
       end
     end
   end
 
   describe "#comment=" do
     it "sets a comment to the record" do
-      comment = "sample comment"
+      comment = "#sample comment"
```
joseivanlopez: Augeas saves what you pass as comment literally. I agree, it is necessary to indicate it.

src/lib/cfa/ntp_conf.rb
joseivanlopez: Yes, all subclasses implement it. CommandRecord implements it for all its derived classes (ServerRecord, FudgeRecord, etc), which share the same logic for this method.

src/lib/cfa/ntp_conf.rb
joseivanlopez: Yes, your are right. I correct it.

src/lib/cfa/ntp_conf.rb
joseivanlopez: You are righth. I change it.

test/cfa/ntp_conf_test.rb
```diff
@@ -39,36 +47,80 @@ def ntp_disk_file
...
     ntp.load
   end
 
-  context "#add" do
+  context "#<<" do
     context "when does not exist the record to add" do
       it "adds the record" do
-        ntp.records.add(new_record)
+        ntp.records << new_record
         expect(ntp.records.include?(new_record)).to be(true)
```
joseivanlopez: I agree.

test/cfa/ntp_conf_test.rb
joseivanlopez: Right now it is not possible because it entails several changes. Augeas structure is sligtly different for these two cases. Anyway, ntp-client does not allow to change restrict entries currently.

src/lib/cfa/ntp_conf.rb
joseivanlopez: So, in this case, I understand that it better to declare that methods as protected.

src/lib/cfa/ntp_conf.rb
```diff
@@ -0,0 +1,390 @@
...
+        entry[:key] << "[]" if COLLECTION_KEYS.include?(entry[:key])
+        fix_keys(entry[:value].tree) if entry[:value].is_a?(AugeasTreeValue)
+      end
+    end
+
+    # class to manage ntp entries as a collection.
+    #
+    # Each element of the collection is a subclass of Record.
+    # A Record object is a wrapper of an AugeasElement.
+    class RecordCollection
```
jreidinger: I just notice one thing. this class is in private section, but in fact it is public API as above method records return it, so it should be also properly documented, especially public API as user of `records` method need to know it.

src/modules/NtpClient.rb
```diff
@@ -282,6 +296,38 @@ def GetNtpServersByCountry(country, terse_output)
       end
     end
 
+    def read_ntp_conf
+      if !File.exist?(NTP_FILE)
```
jreidinger: one question. If it is used in installation, it will read file from insts-sys and not target system, which we probably do not want. In such case I would like to point you to two yast classes - https://github.com/yast/yast-yast2/blob/5762181d62762816a73fc040362c1efb5d97deed/library/general/src/modules/FileUtils.rb#L54 and https://github.com/yast/yast-yast2/blob/5762181d62762816a73fc040362c1efb5d97deed/library/system/src/lib/yast2/target_file.rb#L54

src/modules/NtpClient.rb
```diff
@@ -1062,22 +1082,37 @@ def new_write_progress
...
+          record["cfa_record"] = CFA::NtpConf::Record.record_class(record["type"]).new
+          @ntp_conf.records << record["cfa_record"]
+        end
+        update_cfa_record(record)
+      end
 
-      return false if !SCR.Write(path(".etc.ntp_conf.all"), conf)
+      @deleted_records.each do |cfa_record|
+        @ntp_conf.records.delete(cfa_record)
+      end
```
jreidinger: you can implement it with delete_if `@ntp_conf.records.delete_if { |r| @deleted_records.include?(r) }`

src/modules/NtpClient.rb
```diff
@@ -1062,22 +1082,37 @@ def new_write_progress
...
 
-      ret = SCR.Write(path(".etc.ntp_conf"), nil)
+      begin
+        @ntp_conf.save
+      rescue
+        return false
```
jreidinger: if save failed at least log why

test/ntp_client_test.rb
```diff
@@ -646,49 +637,33 @@
...
     it "returns false if config has been read previously" do
       subject.instance_variable_set(:@config_has_been_read, true)
-
       expect(subject.ProcessNtpConf).to eql(false)
     end
 
     it "returns false if config doesn't exist" do
-      allow(Yast::FileUtils).to receive(:Exists).with("/etc/ntp.conf").and_return(false)
-
+      allow(File).to receive(:exist?).and_return(false)
```
jreidinger: difference is that original FileUtils in insts-sys checks in installation environment when file exist on target disk

src/modules/NtpClient.rb
joseivanlopez: Ok. I restore it.

src/modules/NtpClient.rb
joseivanlopez: Done

src/lib/cfa/ntp_conf.rb
```diff
@@ -99,15 +99,27 @@ def fix_keys(tree)
...
+    # object for each line. In this case, two
+    # ServerRecord and a PeerRecord.
     class RecordCollection
       include Enumerable
 
       def initialize(augeas_tree)
         @augeas_tree = augeas_tree
       end
 
+      # Iterates the elements of the collection.
```
jreidinger: maybe add something like `@see Array#each`

jreidinger: and add what parameter yield have http://www.rubydoc.info/gems/yard/file/docs/Tags.md#yield

src/lib/cfa/ntp_conf.rb
```diff
@@ -130,6 +142,7 @@ def delete(record)
         reset_cache
       end
 
+      # Removes all records that satisfy a condition.
```
jreidinger: document that yield get Record class see http://www.rubydoc.info/gems/yard/file/docs/Tags.md#yield

src/lib/cfa/ntp_conf.rb
joseivanlopez: Done

src/lib/cfa/ntp_conf.rb
joseivanlopez: Done

src/lib/cfa/ntp_conf.rb
```diff
@@ -119,7 +119,8 @@ def initialize(augeas_tree)
         @augeas_tree = augeas_tree
       end
 
-      # Iterates the elements of the collection.
+      # Iterates the elements of the collection (@see Array#each).
+      # @yield [record] gives a Record of the collection
```
jreidinger: type should be class. so `[Record]` so yardoc can link it.

package/yast2-ntp-client.spec
```diff
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-ntp-client
-Version:        3.2.1
+Version:        3.2.2
```
jreidinger: that reminds me, one important thing is missing here. Dependency on cfa gem and also with correct version that have your changes.

PR #66, opened 2016-12-15
Fixed some tests (bsc#983486).

teclator:
Just fixed some tests that does not allow to build the package correctly. Travis does not complain because it uses the old ntp.lense

Josef Reidinger fixed the ntp lense in this commit:

hercules-team/augeas@01e4c42

Basically before the fix it doesn't expect to have restrict records with IPv4 explicitly, as for example:

`restrict -4 default`

So before the fix `-4` was treated as an **address** and `default` as an **option* what is wrong.

With the new lense, **cfa** will read the **IP version** correctly.

**IN PROCCESS**
```
The ubuntu repository for augeas is ppa:raphink/augeas, which installs:
Setting up augeas-lenses (1.4.0-0ubuntu1~augeas1~precise4) ...
Setting up libaugeas0 (1.4.0-0ubuntu1~augeas1~precise4) ...
```
We need updated packages.

### DIFF
--- a/package/yast2-ntp-client.spec
+++ b/package/yast2-ntp-client.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-ntp-client
-Version:        3.2.2
+Version:        3.2.3
 Release:        0
 Summary:        YaST2 - NTP Client Configuration
 License:        GPL-2.0+
@@ -65,7 +65,7 @@ rake install DESTDIR="%{buildroot}"
 %files
 %defattr(-,root,root)
 %dir %{yast_yncludedir}/ntp-client
-%{yast_dir}/clients/*
+%{yast_clientdir}/*
 %{yast_dir}/lib
 %{yast_yncludedir}/ntp-client/*
 %{yast_scrconfdir}/cfg_ntp.scr
--- a/package/yast2-ntp-client.spec
+++ b/package/yast2-ntp-client.spec
@@ -65,9 +65,9 @@ rake install DESTDIR="%{buildroot}"
 %files
 %defattr(-,root,root)
 %dir %{yast_yncludedir}/ntp-client
+%{yast_dir}/clients/*
+%{yast_dir}/lib
 %{yast_yncludedir}/ntp-client/*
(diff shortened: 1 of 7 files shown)


test/cfa/ntp_conf_test.rb
```diff
@@ -359,13 +359,13 @@ def ntp_conf(file)
 
   describe "#options" do
     it "obtains the options of the record" do
-      expect(record.options).to eq(%w(default notrap nomodify nopeer))
```
teclator: **default** is not an option but an address

test/ntp_client_test.rb
```diff
@@ -679,7 +683,7 @@
     end
 
     it "initializes restrict records" do
-      expect(subject.restrict_map.size).to eql(4)
+      expect(subject.restrict_map.size).to eql(3)
```
teclator: Basically it only reads 3 restrict records because they are indexed by the "address" and 
```
restrict -4 default
restrict -6 default
```
are indexed by the "default" key. With the older lense there was an error which added "-4" as an address that is the reason of having 4 entries previously.

package/yast2-ntp-client.spec
```diff
@@ -65,9 +65,9 @@ rake install DESTDIR="%{buildroot}"
 %files
 %defattr(-,root,root)
 %dir %{yast_yncludedir}/ntp-client
+%{yast_dir}/clients/*
```
jreidinger: it can be `%{yast_clientdir}/*`

jreidinger:
LGTM. regarding failing travis only option I see is to install somehow new lense ( from some ubuntu repo or manually ) or wait when @lslezak POC with docker will be used more widely.

package/yast2-ntp-client.spec
teclator: Updated

lslezak:
It's green now! The Docker rocks! :+1:

teclator:
Cool!!, thnx @lslezak for make our lives better!! well at least more green ;)

and thnx @jreidinger for review!!

PR #67, opened 2016-12-30
AY: do not export Augeas objects

schubi2:


src/modules/NtpClient.rb
```diff
@@ -697,17 +697,23 @@ def Export
       # restrict_map is a map with the key ip,ipv4-tag or ipv6-tag.
       # This will be converted into a list in order to use it in
       # autoyast XML file properly.
-      restricts = @restrict_map.collect do |target, values|
+      restricts = deep_copy(@restrict_map)
```
ancorgs: I would try to not introduce more usage of `deep_copy` (is one of those leftovers from ycp -> ruby migration we would like to die).

What about?
```ruby
restrics = @restrict_map.collect do |target, values|
  # cfa_record not needed for export
  export_values = values.dup.delete("cfa_record")
  export_values["target"] = target
  export_values
end

peers = @ntp_records.map do |peer|
  # cfa_record not needed for export
   peer.dup.delete("cfa_record")
end
```
It's the same but without "ruby aliens"

test/ntp_client_test.rb
```diff
@@ -21,6 +22,20 @@
...
+    before(:each) do
+      Yast::Profile.ReadXML(profile)
+    end
+
+    it "imports/exports settings from/to AutoYaST configuration file" do
+      subject.Import(Yast::Profile.current["ntp-client"])
+      expect(subject.Export()).to eql(Yast::Profile.current["ntp-client"])
+    end
+  end
+
```
ancorgs: You are merging the test for `#Import` and `#Export`into a single `it` and using too generic description for my particular taste. Now that we are not in a hurry, may I suggest more fine-grained tests? Like, for example.

```ruby
describe "AutoYaST methods" do
  FIXTURES_PATH = File.join(File.dirname(__FILE__), 'fixtures')
  let(:ntp_client_section) do
    file = File.join(fixtures_path, 'autoyast', profile_name) }
    Yast::Profile.ReadXML(profile)
    Yast::Profile.current["ntp-client"]
  end

  describe "#Import" do
    context "with a correct AutoYaST configuration file" do
      let(:profile_name) { "autoinst.xml" }

      it "reads the list of peers" do
        subject.Import(ntp_client_section)
        expect(subject.ntp_records.size).to eq 4
      end

      it "reads the list of restricts" do
        subject.Import(ntp_client_section)
        expect(subject.restrict_map.keys).to contain_exactly(
          "default", "127.0.0.1", "::1"
        )
      end
    end

    context "with an invalid AutoYaST configuration file" do
      let(:profile_name) { "wrong.xml" }

      it "whatever" do
        pending # Not sure what should happen here
      end
    end

    context "with an empty AutoYaST configuration" do
      let(:ntp_client_section) { {} }

      it "whatever" do
        pending # Not sure what should happen here
      end
    end
  end

  describe "#Export" do
    let(:profile_name) { "autoinst.xml" }

    it "produces an output equivalent to #Import" do
      subject.Import(ntp_client_section)
      expect(subject.Export()).to eq ntp_client_section
    end
  end
end
```

PR #71, opened 2017-03-21
Adapt new cfa

jreidinger:
https://trello.com/c/qLa4HuV9/855-8-caasp-fix-zypp-conf-modifications


mvidner:
Huh, there is more CFA code than I expected. Did you try a basic interactive test of this?

jreidinger:
@mvidner yes, I try rake run and some clicking and it works. But when I try add new server then after new start nothing happen ( it is not there ). Not sure if problem was there even before or not.

test/cfa/ntp_conf_test.rb
```diff
@@ -59,9 +59,11 @@ def ntp_conf(file)
       it "writes a new entry" do
         record = CFA::NtpConf::ServerRecord.new
         record.value = "3.pool.ntp.org"
+        record.raw_options = "iburst"
+        record.comment = nil
```
mvidner: Why is this comment=nil needed?

src/lib/cfa/ntp_conf.rb
```diff
@@ -125,10 +125,17 @@ def each(&block)
         record_entries.each(&block)
       end
 
+      # Get last record in collection
+      def last
+        record_entries.last
+      end
+
       # Adds a new Record object to the collection.
+      # @note it creates new instance of record after this.
```
mvidner: "after"? I don't understand

src/modules/NtpClient.rb
```diff
@@ -1099,10 +1099,12 @@ def update_cfa_record(record)
     def write_ntp_conf
       records_for_write.each do |record|
         unless record["cfa_record"]
-          record["cfa_record"] = CFA::NtpConf::Record.record_class(record["type"]).new
-          ntp_conf.records << record["cfa_record"]
+          ntp_conf.records << CFA::NtpConf::Record.record_class(record["type"]).new
+          record["cfa_record"] = ntp_conf.records.last
```
mvidner: We fixed an obscure bug here but the method remains untested? Will it break again once we touch CFA?

test/cfa/ntp_conf_test.rb
jreidinger: not needed, I just test some options there and in real code this happens sometimes.

src/lib/cfa/ntp_conf.rb
jreidinger: it means, that you have instance of Record `r` and you call `records << r` but in fact in records, there is another instance and not `r` itself. So if you modify it later, like it happen in code, it is ignored.

src/modules/NtpClient.rb
jreidinger: well, if you feel brave enough, you can try to write reasonable test for write, which verify that really modified file is written. Now it do a lot of stuff and I do not understand half of them.

test/cfa/ntp_conf_test.rb
mvidner: Please keep the original `it` as is and add a new one with `iburst` and a new one with `comment`.

jreidinger: ok

src/lib/cfa/ntp_conf.rb
jreidinger: maybe I can rephrase it like `@note argument is not member of collection and instead new instance is created from its augeas content`

mvidner: yes, thanks

src/modules/NtpClient.rb
mvidner: Yet you felt brave enough to change CFA incompatibly, and [it will happen again for 1.0](https://github.com/config-files-api/config_files_api/issues/17). There was a bug not covered with tests that you had to find by manually running the module. Come on, write the test!

jreidinger: well, if there is a rule to write test for any code you touch or fix, I worry we should know it in advance before planning meeting, as we are quite a lot underestimated

jreidinger:
currently it failing because:

1. new yast2 is not yet in docker image
2. ntpclient test is disaster, any modification done to module is not reset after test.

jreidinger:
well, I fix second problem and now only need to check that with new yast2 it will pass

PR #73, opened 2017-03-23
one more fix for backward compatibility

jreidinger:


teclator:
@jreidinger rubocop complains :(. Maybe as it's for backward compatibility could be useful to use the old ntp lens to be sure that it also pass the tests only for the restrict ones.

jreidinger:
@teclator it is in fact tested with obs build, which use respective augeas version, so if it failed when build against SP3, it is wrong. I think it do not make much sense to test it against lens that is not used in given product.

teclator:
@jreidinger of course but in that case it does not needs backward compatibility... ;)

PR #74, opened 2017-04-21
Add CFA class

imobachgs:
NtpClient does not handle the configuration file properly. The problem is fixed in `master`, but we need to fix also for CaaSP. So I've decided to port (actually to copy) the CFA class (used in master).

This class won't be used by `NtpClient` as there are too many changes involved and I would like to be as conservative as possible at this point.

Of course, CaaSP version will diverge from SLE 12 SP2.


imobachgs:
Thanks!

PR #75, opened 2017-04-24
Add a test case about writing more than 1 server with options

imobachgs:


imobachgs:
This PR was just to demonstrate a problem in CFA, so now that [it's fixed](https://github.com/config-files-api/config_files_api/pull/19), we can close it.

PR #77, opened 2017-05-31
added some more tags for ntp.conf

schubi2:


src/lib/cfa/ntp_conf.rb
```diff
@@ -231,6 +254,7 @@ def value=(value)
       end
 
       def comment
+        return augeas[:mutline] if augeas[:multiline]
```
joseivanlopez: Typo: `augeas[:mutline]`. Are tests strong enough?

src/lib/cfa/ntp_conf.rb
```diff
@@ -70,6 +76,20 @@ def load
       fix_keys(data)
     end
 
+    def save
```
joseivanlopez: Please, add doc to this new method.

src/lib/cfa/ntp_conf.rb
```diff
@@ -137,6 +157,8 @@ def last
       # @param [Record] record
       def <<(record)
         @augeas_tree.add(record.augeas[:key], record.augeas[:value])
+        # TODO: nasty workaround to survive multiline key with long comments
+        @augeas_tree.all_data.last[:multiline] = record.augeas[:multiline]
```
joseivanlopez: Is it `:multiline` a real augeas key? Is it not possible to use `comment[]` for multiline comments?

test/cfa/ntp_conf_test.rb
```diff
@@ -80,7 +80,18 @@ def ntp_conf(file)
...
+      end
 
+      it "can write multi lines comments from autoyast profiles" do
+        record = CFA::NtpConf::ServerRecord.new
+        record.value = "3.pool.ntp.org"
+        record.comment = "# test comment\n#test comment 2\n#test comment3"
+        ntp.records << record
+        ntp.save
+        expect(file.content.lines).to include("server 3.pool.ntp.org\n")
+        expect(file.content.lines).to include("## test comment\n")
```
joseivanlopez: Why have the resulting lines two `#` ?

jreidinger: looks like problem with lense, I have to inspect it a bit more.

src/lib/cfa/ntp_conf.rb
jreidinger: good point, I do not test comment for multiline.

src/lib/cfa/ntp_conf.rb
jreidinger: No, multiline is not real key and I use it only to overwrite it above in save method. it is possible, but not with current implementation of record, where record is one augeas element. and these comments will be siblings. So it is a bit work-around only for autoyast.

test/cfa/ntp_conf_test.rb
```diff
@@ -407,3 +419,21 @@ def ntp_conf(file)
...
+
+  it "adds proper type to conf when saving" do
+    ntp.records << subject
+    record = ntp.records.last
+    record.value = "1"
+    record.raw_options = ""
+    record.comment = "# path for keys file\n"
+    expect(record.value).to eq "1"
+
+    ntp.save
```
joseivanlopez: Please, check that correct value is written to file. In this case, I guess:

`expect(file.content.lines).to include("trustedkey 1# path to keys file")`

test/cfa/ntp_conf_test.rb
jreidinger: well, it really depends on how lense is written. There are lenses which have "#" matched and stored, so it is part of final string, and some that just require it, but do not store it, so it is not part of final string. It is a bit inconsistent there. And even more challenging in files that allow both "#" and ";" as comment starter.

test/cfa/ntp_conf_test.rb
jreidinger: it is actually a bit funny, as `record.comment = "# path for keys file\n"` result in different result then if I use `record.comment = "# path for keys file"` The first one result in two lines one, as comment include endline. The second one do what is expected as trailing one.

PR #80, opened 2017-06-26
Fix bug creating restrict records

joseivanlopez:
PBI: https://trello.com/c/7M590k9O/1045-3-sles12-sp3-p1-build-0439-autoyast-installation-crashes-in-ntp

Bug: https://bugzilla.suse.com/show_bug.cgi?id=1045777

### DIFF
--- a/package/yast2-ntp-client.changes
+++ b/package/yast2-ntp-client.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Jun 26 17:23:52 UTC 2017 - jlopez@suse.com
+
+- AutoYaST: Fixed crash in CFA when creating restrict record from
+  scratch (bsc#1045777)
+- 3.2.9
+
+-------------------------------------------------------------------
 Fri Jun  2 16:37:27 CEST 2017 - schubi@suse.de
 
 - AutoYaST: Fixed crash in CFA class while writing settings to
--- a/src/lib/cfa/ntp_conf.rb
+++ b/src/lib/cfa/ntp_conf.rb
@@ -488,8 +488,8 @@ module CFA
     # For example:
     #   restrict -4 default notrap nomodify nopeer noquery
     #
-    # FIXME: right now there is not way to create an ipv4 restrict
-    #   record from scratch using old lense. The reason is that it
+    # FIXME: right now there is no way to create an ipv4 restrict
+    #   record from scratch using old lens. The reason is that it
     #   is not possible to detect an old_lens? case for a new record.
     #
     # Restrict entry has its own options interpretation.
--- a/src/lib/cfa/ntp_conf.rb
+++ b/src/lib/cfa/ntp_conf.rb
@@ -488,18 +488,31 @@ module CFA
     # For example:
     #   restrict -4 default notrap nomodify nopeer noquery
     #
+    # FIXME: right now there is not way to create an ipv4 restrict
+    #   record from scratch using old lense. The reason is that it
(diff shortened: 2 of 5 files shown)


test/cfa/ntp_conf_test.rb
```diff
@@ -418,6 +418,17 @@ def ntp_conf(file)
       expect(file.content).to include("restrict -4 default notrap\n")
     end
   end
+
+  context "when creating a new record from scratch" do
+    it "sets values correctly" do
+      record = CFA::NtpConf::RestrictRecord.new
+      record.value = "192.168.0.1"
+      record.options = ["default"]
```
teclator: I complain :) default is not an option but an address, it is used when the ip protocol version is used just because the augeas lens in use does not covers -4 option, but in this case as you are using an IP address then I prefer to use real options like `nomodify notrap nopeer noquery`

package/yast2-ntp-client.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Jun 26 17:23:52 UTC 2017 - jlopez@suse.com
+
+- AutoYaST: Fixed crash in CFA when creating restrict record from
+  scratch (bsc#1045777)
+- 3.2.9
```
teclator: Mm should we move to 3.3.0 in SP3 or no?

jreidinger: no, 3.3.0 make sense only in master

teclator: great

src/modules/NtpClient.rb
```diff
@@ -1090,7 +1090,7 @@ def new_write_progress
 
     def update_cfa_record(record)
       cfa_record = record["cfa_record"]
-      cfa_record.value = record["address"]
+      cfa_record.value = record["address"].strip
```
jreidinger: I think it should be handled even more upper level, record should not contain newline. so probably somewhere near autoyast import.

src/lib/cfa/ntp_conf.rb
```diff
@@ -488,37 +488,54 @@ def raw_options=(raw_options)
     # For example:
     #   restrict -4 default notrap nomodify nopeer noquery
     #
+    # FIXME: right now there is not way to create an ipv4 restrict
```
jreidinger: typo "there is no way"

src/modules/NtpClient.rb
```diff
@@ -676,6 +676,9 @@ def Import(settings)
         end
         next deep_copy(p)
       end
+
+      @ntp_records.each { |record| record["address"].strip! }
```
jreidinger: why not others? e.g. type cannot have whitespaces after it? and what about options?

joseivanlopez: Then, I would do it for all keys, right?

imobachgs: np: why no `@ntp_records.each { |r| r["address"].strip! }` ?

jreidinger: yes, make sense to do it for all input.

src/modules/NtpClient.rb
```diff
@@ -986,6 +987,13 @@ def AutoPackages
 
   private
 
+    # Remove blank spaces in values
+    def sanitize_record(record)
+      sanitized = record.dup
+      sanitized.each_key { |k| sanitized[k].strip! }
```
jreidinger: why not `sanitized.each_value { |v| v.strip! }` ?

PR #83, opened 2017-06-29
Take in account key value options in command records (bsc#1045968)

teclator:


src/lib/cfa/ntp_conf.rb
```diff
@@ -346,13 +356,32 @@ def options_matcher
...
+
+        # Some options can be a key value pair or just a single one
+        options.each do |option|
+          if KEY_VALUE_CMD_OPTIONS.include?(option)
+            key = option
+          else
+            add_option(key, option)
+            key = nil
+          end
+        end
```
jreidinger: I think this can be problem if value for key-value one is same as some key like `mode ident version 3`. This result in `{ "ident" => "version"}` result.

So maybe code like this
```
i = 0
while i < options.size
  if KEY_VALUE_CMD_OPTIONS.include?(options[i])
    add_options(options[i], options[i] + 1)
    i += 2
  else
   add_options(nil, options[i])
   i += 1
  end
end
```
which I find also a bit more readable.

jreidinger: or even 
```
i = 0
while i < options.size
  if KEY_VALUE_CMD_OPTIONS.include?(options[i])
    add_options(options[i], options[i] + 1)
    i += 2
  else
   add_options(options[i], nil)
   i += 1
  end
end
```
and modify add_option to
```
      def add_option(key, option)
        tree_value.tree.add(key, option)
       end
```

teclator: @jreidinger this was the first option I was considering but according to documentation:

https://www.eecis.udel.edu/~mills/ntp/html/confopt.html#option

It should not be a valid entry. 

But if you consider it more readable, then for me is enough to rewrite it ;)

jreidinger: well, this trick that make passing it as value and having key nil, then later in `add_option` switch key and value if key is nil is really magic for me

teclator: @jreidinger updated

PR #85, opened 2017-06-30
Fix bug sanitizing values

joseivanlopez:


package/yast2-ntp-client.changes
```diff
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Fri Jun 30 09:06:48 UTC 2017 - jlopez@suse.com
+
+- AutoYaST: Fix crash when sanitizing values
```
teclator: we could explain exactly which values, and at least modify this profile adding the 

```<initial_sync config:type="boolean">true</initial_sync>```

to the first peer in:

test/fixtures/autoyast/autoinst.xml

**Comment by teclator** [comment]
We have a problem with the entries that have a path or file need a trailing space. Like:

`driftile`, `includefile`, `leapfile`, `logfile`, `saveconfigdir`, `crypto`, `keys`, `keysdir` and probably `ident`

https://www.eecis.udel.edu/~mills/ntp/html/miscopt.html
https://www.eecis.udel.edu/~mills/ntp/html/authopt.html

@schubi2 is currently testing a profile with a section like:

```
      <peer>
        <address>/etc/ntp.keys    </address>
        <options/>
        <type>keys</type>
        <comment># Keys file
</comment>
      </peer>
```

And I expect it to fail because we are removing the spaces from the address and then the entry is written like:

```keys /etc/ntp.keys# Keys file```

which looks fine for augeas lense but ntpd then expects ```/etc/ntp.keys#``` instead of ```/etc/ntp.keys``` and the service will not start.

**Comment by joseivanlopez** [comment]
@teclator I think the scenario you propose works fine with our code because comment has a break line. The following won't work, I guess.

```
      <peer>
        <address>/etc/ntp.keys    </address>
        <options/>
        <type>keys</type>
        <comment># Keys file</comment>
      </peer>
```

**Comment by teclator** [comment]
@joseivanlopez true, this one is the example I wanted to describe the other one will be written as

```
# Keys file
keys /etc/ntp.keys
```

File: package/yast2-ntp-client.changes [hunk 2]
```diff
@@ -1,4 +1,13 @@
 -------------------------------------------------------------------
+Fri Jun 30 09:36:48 UTC 2017 - jlopez@suse.com
+
+- AutoYaST: Fix crash when trying to sanitize not string values,
+  for example initial_sync.
+- AutoYaST: Sanitize comments by only removing blank spaces at
+  beginning and adding line break. This avoids augeas parse errors.
+- 3.2.11
```
teclator: bug number is missing

PR #86, opened 2017-06-30
Write restrict records first (bsc#983486)

teclator:
Trello card: https://trello.com/c/N9tffDE4/1062-sles12-sp2-p3-983486-yast2-ntp-client-mangles-etc-ntpconf

Default template is something like:

```
# Comment
# Comment
# Comment
restrict -4 default notrap...
restrict -6 default notrap...

pool.ntp.example.org
```

And after read and write it was written as:

```
pool.ntp.example.org
# Comment
# Comment
# Comment
restrict -4 default notrap...
restrict -6 default notrap...
```

This PR just writes the restrict entries before the rest.

### DIFF
--- a/src/modules/NtpClient.rb
+++ b/src/modules/NtpClient.rb
@@ -598,6 +598,8 @@ module Yast
       # write settings
       return false if !go_next
 
+      # Restrict map records are written first to not mangle the config file
+      # (bsc#983486)
       @ntp_records = restrict_map_records + @ntp_records
 
       log.info "Writing settings #{@ntp_records}"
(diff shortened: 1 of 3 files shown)


src/modules/NtpClient.rb
```diff
@@ -598,7 +598,9 @@ def Write
       # write settings
       return false if !go_next
 
-      @ntp_records += restrict_map_records
+      # Restrict map records are written first to not mangle the config file
+      # (bsc#983486)
+      @ntp_records = restrict_map_records + @ntp_records
```
teclator: I don't like that it adds restrict maps to ntp_records each time Write is called, we probably could remove this line and just concatenate both arrays in records_for_write, but is preferable to do it in a separate PR.

PR #93, opened 2017-08-03
Backport crash fix

jreidinger:


teclator:
It basically backports  #93

lslezak:
BTW yet another example which could have been caught by https://github.com/yast/yast-rear/pull/8...

jreidinger:
@lslezak well, it expect some clicking to advanced settings, so I still worry we are getting into area of openQA

lslezak:
But as you can see it is not handled in openQA at all...

jreidinger:
@lslezak yes, as test missing in openQA. So we have option to write it in openQA or create new framework and write it there. To be honest I am not sure if own framework is better option.

PR #99, opened 2017-11-03
Chrony support

jreidinger:
https://trello.com/c/bXvReOld/1261-8-yast-ntp-client-working-on-a-running-system


src/clients/ntp-client_auto.rb
```diff
@@ -63,32 +58,17 @@ def main
...
       elsif @func == "Change"
-        @ret = NtpClientAutoSequence()
+        # TODO: implement
+        Yast::Report.Error("Not supported yet for chrony")
+        @ret = :next
       # Import configuration
       elsif @func == "Import"
-        @ret = NtpClient.Import(@param)
+        Yast::Report.Error("Not supported yet for chrony")
+        @ret = true
```
mchf: hm, shouldn't it return false (popups can be timed out in some scenarios)

jreidinger: no, requirement was to not crash, so it should not complain now ( beside that error message ).

src/clients/ntp-client_proposal.rb
```diff
@@ -387,27 +376,13 @@ def Write(param)
...
-        )
-        Builtins.y2milestone("'sntp %1' returned %2", ntp_server, ret)
+        NtpClient.sync_once(ntp_server)
         Popup.ClearFeedback
       end
 
       return :ntpdate_failed if ret != 0
 
-      # User wants to more than running sntp (synchronize on boot)
+      # User wants to more than running one time sync (synchronize on boot)
```
mchf: I'd use ```wants more than``` instead of ```wants to more```

jreidinger: right, I just change second part of sentence, but first part is wrong from begging. Will fix it.

src/clients/ntp-client_proposal.rb
```diff
@@ -315,10 +300,13 @@ def AskUser
     def WriteNtpSettings(ntp_servers, ntp_server, run_service)
```
mchf: this function has interesting parameters which deserve some documentation. E.g. if i understand it correctly ```ntp_server``` is ignored when ```ntp_servers``` is not empty.

src/clients/ntp-client_proposal.rb
```diff
@@ -315,10 +300,13 @@ def AskUser
...
       ntp_servers = deep_copy(ntp_servers)
       NtpClient.modified = true
+      NtpClient.ntp_conf.clear_pools
       if ntp_servers != []
```
mchf: btw
```
ntp_servers << [ntp_server] if !ntp_servers.empty?
```
should work the same and drop the strange if-else branch

src/clients/ntp-client_proposal.rb
jreidinger: yes, old code, but will document it.

src/clients/ntp-client_proposal.rb
```diff
@@ -349,6 +337,7 @@ def WriteNtpSettings(ntp_servers, ntp_server, run_service)
     # return:
     #   `success, `invalid_hostname or `ntpdate_failed
     def Write(param)
+      Builtins.y2milestone("ntp client proposal Write with #{param.inspect}")
```
mchf: ```Bultins``` in new code?

src/clients/ntp-client_proposal.rb
jreidinger: hmm, I do not get your code, but probably something like `ntp_servers << ntp_server if ntp_servers.empty?` will work.

src/clients/ntp-client_proposal.rb
jreidinger: will modify it.

src/clients/ntp-client_proposal.rb
```diff
@@ -362,10 +351,10 @@ def Write(param)
       return :success if param["write_only"]
 
       # One-time adjusment without running the ntp daemon
-      # Meanwhile, ntpdate was replaced by sntp
+      # Meanwhile, ntpdate was replaced by sntp and later by chrony
```
mchf: well it is nice to know what ntp tools used to be used in suse but it belongs into chronicle not here

src/clients/ntp-client_proposal.rb
```diff
@@ -362,10 +351,10 @@ def Write(param)
...
       ntpdate_only = param["ntpdate_only"]
```
mchf: I'd remove this assignment. Its name is misleading. It is used like twenty lines lower and at one place only.

src/clients/ntp-client_proposal.rb
jreidinger: fair enough, will remove it.

src/clients/ntp-client_proposal.rb
```diff
@@ -387,27 +376,13 @@ def Write(param)
...
-        Builtins.y2milestone("'sntp %1' returned %2", ntp_server, ret)
+        NtpClient.sync_once(ntp_server)
         Popup.ClearFeedback
       end
 
       return :ntpdate_failed if ret != 0
 
-      # User wants to more than running sntp (synchronize on boot)
+      # User wants more than running one time sync (synchronize on boot)
       WriteNtpSettings(ntp_servers, ntp_server, run_service) if !ntpdate_only
```
mchf: here it is so ```ntpdate_only``` means "synce once only"

src/lib/cfa/chrony_conf.rb
```diff
@@ -0,0 +1,142 @@
...
+    end
+
+    # loads cfa model and ensure all collection keys have [] suffix
+    def load
+      super
+      fix_collection_names(data)
+    end
+
+    # options hash with key value, where value can be nil for keyword options
+    # or :default to use default options
```
mchf: ok and what is ```address``` param good for?

jreidinger: will add properly both params.

src/lib/cfa/chrony_conf.rb
```diff
@@ -0,0 +1,142 @@
...
+    def add_pool(address, options = :default)
+      options = default_pool_options if options == :default
+      # if there is already pool entry, place it after, if not, try use comment
+      existing_pools = pure_pools
+      if existing_pools.empty?
+        # for now first chrony have pools under comment mentioning pool.ntp.org
+        # so try to place it below
+        matcher = Matcher.new { |k, v| k.start_with?("#comment") && v =~ /www\.pool\.ntp\.org/ }
```
mchf: so if user removes the comment we fail?

jreidinger: no, AfterPlacer if matcher does not find entry, append it to end of file. So strategy is:

1. if there is any pool, place it afterwards
2. place after comment for pools
3. append to end of file

mchf: no ... just add at the end of conf file (tested)

jreidinger: I will document that behavior about where line is placed.

src/clients/ntp-client_proposal.rb
mchf: hm yes that's it  ... square brackets were bonus for free :-/

src/lib/y2ntp_client/widgets/main_widgets.rb
```diff
@@ -0,0 +1,254 @@
...
+
+        items
+      end
+
+      def init
+        self.value = Yast::NtpClient.ntp_policy
+      end
+
+      def store
+        return if value == Yast::NtpClient.ntp_policy
```
mchf: missing at least basic input validation

jreidinger: do we have any helper for it? I am not sure what everything netconfig supports.

src/lib/y2ntp_client/widgets/main_widgets.rb
```diff
@@ -0,0 +1,254 @@
...
+      def opt
+        [:notify]
+      end
+
+      def items
+        [
+          # radio button
+          ["never", _("Only &Manually")],
+          # radio button
+          ["sync", _("&Synchronize without Daemon")],
```
mchf: something like "Synchronize periodically" sounds better to me.

jreidinger: @mvidner what do you think? this is original naming, so we should change it if benefits comes over confusion of existing users.

src/lib/y2ntp_client/widgets/main_widgets.rb
mchf: I'm not aware of such helper. Basically the variable can contain space separated list of device regexps (like eth*) and so. Definitely it should not contain characters like " or ' as these can break whole config file.

package/yast2-ntp-client.spec
```diff
@@ -39,11 +39,11 @@ BuildRequires:  rubygem(%rb_default_ruby_abi:yast-rake)
 Requires:       augeas-lenses
 Requires:       yast2 >= 3.2.21
 Requires:       yast2-country-data
+# needed for network/config agent
+Requires:       yast2-network
```
mchf: as it seems we should move the agent into yast2.rpm

jreidinger: :+1:

src/clients/ntp-client_proposal.rb
```diff
@@ -304,21 +288,27 @@ def AskUser
         ntp_server2 = Convert.to_string(
           UI.QueryWidget(Id(:ntp_address), :Value)
         )
-        AddSingleServer(ntp_server2)
+        NtpClient.ntp_conf.clear_pools
+        NtpClient.ntp_conf.add_pool(ntp_server2)
```
mchf: what is difference between ```ntp_server``` and ```ntp_server2``` ?

jreidinger: none, will fix it

package/yast2-ntp-client.spec
jreidinger: but I would like to not slow down this pr, so can be done separately.

src/lib/y2ntp_client/widgets/main_widgets.rb
mvidner: Previously this option was explained in a help text: https://github.com/yast/yast-ntp-client/pull/99/files/d2c1c227b14535ad05eec969c55a89604345fd4f#diff-c9d99695dbc4159f62ce2c189e3c9f08L40 Do we have a help text now?

src/lib/y2ntp_client/widgets/pool_widgets.rb
```diff
@@ -0,0 +1,114 @@
...
+      def store
+        if value
+          @options["offline"] = nil
+        else
+          @options.delete("offline")
+        end
+      end
+
+      def help
+        _("<b>Start Offline</b> specifies if option offline is used. This option skip this " \
```
mvidner: OMG, so much broken English again. VERB FORMS!!

src/lib/y2ntp_client/widgets/main_widgets.rb
jreidinger: @mvidner not yet, but I can add it.

src/lib/y2ntp_client/dialog/main.rb
```diff
@@ -0,0 +1,83 @@
...
+      end
+
+      def abort_button
+        Yast::Label.CancelButton
+      end
+
+      def hardware_clock_widgets
+        if Yast::NtpClient.ntp_conf.hardware_clock?
+          [
+            Label(_("Hardware clock configured as source. YaST will keep it untouched.")),
```
mvidner: (English is OK here, but) I don't understand what this is trying to tell me. So what? Will YaST not take a screwdriver and disassemble my hardware?

jreidinger: Yes, I can guarantee it :). Maybe I should add `will keep this part of configuration untouched.`?

src/lib/y2ntp_client/widgets/main_widgets.rb
```diff
@@ -0,0 +1,263 @@
...
+
+      def label
+        # TRANSLATORS: label for widget that allows to define if ntp configiration is only
+        # from its source or dynamically extended e.g. via DHCP
+        _("Configuration Source")
+      end
+
+      def help
+        # TODO: not written previously, but really deserve something
+        _("<b>Configuration Source</b> defines if network itself can adds ntp sources.")
```
mvidner:
```rb
# TRANSLATORS: 'man 8 netconfig' is a command, do not translate that
_("The NTP configuration may be provided by the local network over DHCP. " \
"<b>Configuration Source</b> can simply enable or disable using that configuration. " \
"In cases where there may be multiple DHCP sources, it can prioritize them: " \
"see 'man 8 netconfig'.")
```

Tip: "can adds" is ALWAYS wrong.

src/lib/y2ntp_client/widgets/main_widgets.rb
```diff
@@ -0,0 +1,273 @@
...
+        items
+      end
+
+      def init
+        self.value = Yast::NtpClient.ntp_policy
+      end
+
+      def validate
+        if value =~ /["']/
+          Yast::Popup.Error(_("Configuration source contain invalid character"))
```
mvidner: The user will hate you for not being told *which* character is invalid :D

```rb
# TRANSLATORS: single quote (') and double quote (") are invalid
_("Configuration Source may not contain single or double quotes")
```

BTW what about SetFocus to the offending widget?

src/lib/y2ntp_client/widgets/pool_widgets.rb
```diff
@@ -0,0 +1,114 @@
...
+          @options["iburst"] = nil
+        else
+          @options.delete("iburst")
+        end
+      end
+
+      def help
+        _("<b>Quick Initial Sync</b> specifies if option iburst is used. This option during " \
+          "initialization send four poll requests with two seconds interval. Useful to quick " \
+          "synchronization during start of machine.")
```
mvidner:
```rb
_("<b>Quick Initial Sync</b> specifies whether the 'iburst' option is used. This option " \
"sends 4 poll requests in 2 second intervals during the initialization. It is useful for a quick " \
"synchronization during the start of the machine.")
```

src/lib/y2ntp_client/widgets/pool_widgets.rb
```diff
@@ -0,0 +1,114 @@
...
+          @options["offline"] = nil
+        else
+          @options.delete("offline")
+        end
+      end
+
+      def help
+        _("<b>Start Offline</b> specifies if option offline is used. This option skip this " \
+          "server during start. It is useful for machine which start without network, because " \
+          "it speed up boot and sync when machine is connected to network.")
```
mvidner:
```rb
_("<b>Start Offline</b> specifies whether the 'offline' option is used. This option skips this " \
"server during the start-up. It is useful for a machine which starts without the network, because " \
"it speeds up the boot, and synchronizes when the machine gets connected to the network.")
```

mvidner:
The translatable strings LGTM!
(As for the rest, I cannot tell, it would take me a couple of hours to review)

PR #101, opened 2017-12-08
Chrony autoyast

jreidinger:
autoyast part for proper export https://github.com/yast/yast-autoinstallation/pull/376


src/autoyast-rnc/ntpclient.rnc
```diff
@@ -4,77 +4,22 @@ namespace config = "http://www.suse.com/1.0/configns"
...
-  element restricts {
-    LIST,
-    restrict*
-  }
-
-start_at_boot =
-  element start_at_boot { BOOLEAN }
-start_in_chroot =
-  element start_in_chroot { BOOLEAN }
+# possible values are systemd, manual and number,
```
mvidner: > Possible values are "systemd", "manual", or a number which means the interval in seconds between synchronizations

An integer or a float?

src/autoyast-rnc/ntpclient.rnc
```diff
@@ -4,77 +4,22 @@ namespace config = "http://www.suse.com/1.0/configns"
...
+ntp_sync = element ntp_sync { text }
 
-sync_interval =
-  element sync_interval { INTEGER }
-synchronize_time =
-  element synchronize_time { BOOLEAN }
+ntp_servers = element ntp_servers {
+      LIST,
+      element ntp_server {
+        element name {text } &
```
mvidner: - `{text }` -> `{ text }` please
- A "name"? I would call it an "address".

src/modules/NtpClient.rb
```diff
@@ -353,18 +366,103 @@ def Write
...
     def Import(settings)
       log.info "Import with #{settings}"
-      # TODO: implement for chrony
+
+      unsupported  = UNSUPPORTED_AUTOYAST_OPTIONS.select { |o| settings.key?(o) }
+      if !unsupported.empty?
+        Yast::Report.Error(
+          format(
+            # TRANSLATORS: error report. %s stands unsuported keys.
+            _("Old unsupported profile detected. Ntp is not configured. Unsupported keys: '%s'."),
```
mvidner: "Ignoring the NTP configuration. The profile format has changed in an incompatible way. These keys are no longer supported: ..."

src/autoyast-rnc/ntpclient.rnc
jreidinger: good note, it is integer. ( float there is rounded )

src/autoyast-rnc/ntpclient.rnc
jreidinger: make sense. It can be ip or hostname address.

src/modules/NtpClient.rb
jreidinger: 👍

src/lib/y2ntp_client/client/auto.rb
```diff
@@ -0,0 +1,75 @@
...
+      def initialize
+        textdomain "ntp-client"
+      end
+
+      def summary
+        Yast::NtpClient.Summary
+      end
+
+      def import(profile)
+        # as current import only modify existing configuration ( and does not contain complete
```
mvidner: English: "As the current Import only modifies ..."

Hm. If this is true, then the `Import({})` below in `reset` does nothing. Is that correct?

jreidinger: no, it do. In fact reset return to previous empty state. So no pools, and default policy and synchronization. Just it does not reread system state. it is questionable if we want reset to system state or to proposed state. But later was done in old days

jreidinger: `Import({})` basically sets internal variable to default values, see e.g. https://github.com/yast/yast-ntp-client/pull/101/files#diff-f31befd8011851c773f783bb9afb356fR383 it sets systemd as default.

PR #102, opened 2017-12-12
Fix error message

imobachgs:
Fixes [bsc#1072407](https://bugzilla.suse.com/show_bug.cgi?id=1072407).

### DIFF
--- a/src/modules/NtpClient.rb
+++ b/src/modules/NtpClient.rb
@@ -371,7 +371,8 @@ module Yast
       if !unsupported.empty?
         Yast::Report.Error(
           format(
-            # TRANSLATORS: error report. %s stands unsuported keys.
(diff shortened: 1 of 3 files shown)


src/modules/NtpClient.rb
```diff
@@ -372,8 +372,9 @@ def Import(settings)
         Yast::Report.Error(
           format(
             # TRANSLATORS: error report. %s stands unsuported keys.
-            _("Ignoring the NTP configuration. The profile format has changed in an " \
-              "incompatible way. These keys are no longer supported: '%s'."),
+            _("Ignoring the NTP configuration. The profile format has changed in an\n" \
```
jreidinger: maybe just mention it in comment for translators, that they also have to fit into such dimenses and if needed, move `\n`

PR #103, opened 2017-12-12
Improve ntp error message

imobachgs:
Improves #102.

wrap-text

### DIFF
--- a/src/modules/NtpClient.rb
+++ b/src/modules/NtpClient.rb
@@ -13,10 +13,12 @@ require "yast"
 require "yaml"
 require "cfa/chrony_conf"
 require "yast2/target_file" # required to cfa work on changed scr
+require "ui/text_helpers"
 
 module Yast
   class NtpClientClass < Module
     include Logger
+    include ::UI::TextHelpers
 
     # the default synchronization interval in minutes when running in the manual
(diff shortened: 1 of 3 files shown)


kobliha:
Nice, thank you! :)

src/modules/NtpClient.rb
```diff
@@ -867,6 +861,20 @@ def pool_servers_for(known_countries)
...
+    def unsupported_error(unsupported)
+      msg = format(
+        # TRANSLATORS: error report. %s stands unsupported keys.
+        _("Ignoring the NTP configuration. The profile format has changed in an " \
+          "incompatible way. These keys are no longer supported: '%s'."),
+        unsupported.join("', '")
+      )
+
+      displayinfo = Yast::UI.GetDisplayInfo
+      width = displayinfo["TextMode"] ? displayinfo.fetch("Width", 80) : 80
```
jreidinger: what about reusing https://github.com/yast/yast-yast2/blob/master/library/general/src/lib/ui/text_helpers.rb#L33 ?

jreidinger: ah, now I see it. So makes sense. Maybe whole this display info changes makes sense to have in text helpers.

PR #104, opened 2017-12-13
Fix import

jreidinger:


package/yast2-ntp-client.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Wed Dec 13 14:40:48 UTC 2017 - jreidinger@suse.com
+
+- Do minimal modification to default chrony.conf in autoyast
+  when chrony is not specified explicitelly in packages section of
```
schubi2: I think it is a typo :-) "explicitly"

src/modules/NtpClient.rb
```diff
@@ -409,6 +409,16 @@ def Import(settings)
       true
     end
 
+    # Merges config to existing system configuration. It is useful for delayed write.
+    # When it at first set values, then chrony is installed and then it writes. So
+    # before write it will merge to system. Result is that it keep majority of config
+    # untouched and modify and what is needed.
```
schubi2: I would remove the last "and"

src/modules/NtpClient.rb
```diff
@@ -409,6 +409,16 @@ def Import(settings)
       true
     end
 
+    # Merges config to existing system configuration. It is useful for delayed write.
+    # When it at first set values, then chrony is installed and then it writes. So
```
schubi2: Hm, I do not really understand this.

jreidinger: I try to explain here that if you set values, it works on parsed configuration file, but if package is not yet installed, then it create new configuration file which is missing many stuff. So if package is installed later, then this method re-apply changes on top of now parsed file.

src/modules/NtpClient.rb
jreidinger: :+1:

PR #105, opened 2018-01-04
[Review] Request from 'schubi2' @ 'yast/yast-ntp-client/review_180104_uid_for_chrony'

schubi2:
Please review the following changes:
  * d2b9114 packaging
  * 81559d8 creating uid if it does not exists

### DIFF
--- a/src/modules/NtpClient.rb
+++ b/src/modules/NtpClient.rb
@@ -153,6 +153,12 @@ module Yast
     def sync_once(server)
       log.info "Running ont time sync with #{server}"
 
+      # E.g. in inst-sys there is no chrony user. We have to
+      # create one manually before calling chronyd.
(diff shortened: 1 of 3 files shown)


src/modules/NtpClient.rb
```diff
@@ -153,6 +153,12 @@ def progress?
     def sync_once(server)
       log.info "Running ont time sync with #{server}"
 
+      # E.g. in inst-sys there is no chrony user. We have to
+      # create one manually before calling chronyd.
```
jreidinger: I still believe it should be handle by chrony itself or installation images, so I prefer if we can wait for @wfeldt

wfeldt: I've created https://github.com/openSUSE/installation-images/pull/223 for this

PR #106, opened 2018-02-02
Bug fixes

jreidinger:


src/lib/y2ntp_client/widgets/main_widgets.rb
```diff
@@ -147,10 +147,14 @@ def label
         _("Synchronization &Interval in Minutes")
       end
 
-      def minimal
+      def minimum
         1
       end
 
+      def maximum
+        59
```
lslezak: NP: Would it be better to define constants for min and max?

jreidinger: yes, but it have to be methods, so I do not see much points to define constant and then use it only in method.

lslezak: I see, then OK.

PR #108, opened 2018-03-07
Conf na,e

jreidinger:


src/modules/NtpClient.rb
```diff
@@ -35,6 +35,9 @@ class NtpClientClass < Module
 
     NTP_FILE = "/etc/chrony.conf".freeze
 
+    # The cron file name for the synchronization.
+    CRON_FILE = "/etc/cron.d/suse-ntp_synchronize"
```
imobachgs: Just add the call to `#freeze`.

lslezak:
:cop: Rubocop complains...

jreidinger:
manual testing of RPM upgrade works well, so I will pleasure rubocop and add changes

package/yast2-ntp-client.changes
```diff
@@ -1,3 +1,13 @@
+-------------------------------------------------------------------
+Wed Mar  7 10:57:36 UTC 2018 - jreidinger@suse.com
+
+- change cron config file name to suse-ntp_synchronize
+  (bsc#1079122)
+- convert old file to new one during upgrade, ensure that ntpd is 
+  replaced by chrony (FATE#323432)
+- ghost that config file, so it is easy to find who own it
```
imobachgs: "who own it" or "which package owns it" ?

imobachgs: or "which package belongs to" ?

src/modules/NtpClient.rb
```diff
@@ -812,13 +812,13 @@ def update_cron_settings
...
           path(".target.string"),
-          @cron_file,
+          CRON_FILE,
           "-*/#{@sync_interval} * * * * root /usr/sbin/chronyd -q &>/dev/null\n"
         )
       else
         SCR.Execute(
           path(".target.bash"),
-          "test -e #{@cron_file} && rm #{@cron_file};"
+          "test -e #{CRON_FILE} && rm #{CRON_FILE};"
```
lslezak: NP: there is a trailing `;` and maybe simple `rm -f #{CRON_FILE}` would be better (even could be replaced by native Ruby)

src/modules/NtpClient.rb
```diff
@@ -812,13 +812,13 @@ def update_cron_settings
       if @synchronize_time
         SCR.Write(
           path(".target.string"),
-          @cron_file,
+          CRON_FILE,
```
kobliha: Just an idea, ... but probably too late: Could we have a comment here that this is created by us and that the user should run YaST to re/configure it?

jreidinger: yes, such comment probably makes good sense. This one is already merged, but I can open new one.

kobliha: Please :)

PR #110, opened 2018-04-04
Add support for Tinker Record (bsc#1086526)

jreidinger:
manual testing needed.

### DIFF
--- a/src/lib/cfa/ntp_conf.rb
+++ b/src/lib/cfa/ntp_conf.rb
@@ -441,7 +441,7 @@ module CFA
 
       # for trustedkey it is subtree of keys
       def value
-        return [] unless tree_value?
+        ensure_tree_value
         key_matcher = CFA::Matcher.new { |k, _v| k == "key" || k == "key[]" }
         keys = tree_value.tree.select(key_matcher)
         keys.map { |option| option[:value] }.join(" ")
@@ -481,7 +481,7 @@ module CFA
       end
 
       def value
-        return [] unless tree_value?
+        ensure_tree_value
         key_matcher = CFA::Matcher.new { |k, _v| k != "#comment" && k != "#comment[]" }
         keys = tree_value.tree.select(key_matcher)
         keys.map { |option| "#{option[:key]} #{option[:value]}" }.join(" ")
--- a/src/lib/cfa/ntp_conf.rb
+++ b/src/lib/cfa/ntp_conf.rb
@@ -40,6 +40,7 @@ module CFA
       manycastclient
       fudge
       restrict
+      tinker
       driftfile
       logfile
       keys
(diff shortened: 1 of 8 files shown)


src/lib/cfa/ntp_conf.rb
```diff
@@ -209,7 +210,9 @@ def reset_cache
       def record_entries
         return @record_entries if @record_entries
         matcher = Matcher.new do |k, _v|
-          RECORD_ENTRIES.include?(k.gsub("[]", ""))
```
teclator: But that would include also not supported keys.. of course CFA will complain. Just thinking out loud..

src/lib/cfa/ntp_conf.rb
```diff
@@ -239,8 +243,22 @@ def self.new_from_augeas(augeas_entry)
       # @param [string] key
       def self.record_class(key)
         entry_type = key.gsub("[]", "")
-        record_class = "::CFA::NtpConf::#{entry_type.capitalize}Record"
-        Kernel.const_get(record_class)
+        record_class = "#{entry_type.capitalize}Record"
+        if CFA::NtpConf.constants.include?(record_class.to_sym)
+          CFA::NtpConf.const_get(record_class.to_sym)
+        else
+          # it is not predefined record type, so use generic misc record
```
teclator: NP: what about *it is not `a` predefined record type, so `a` generic misc record is used*

src/lib/cfa/ntp_conf.rb
jreidinger: we now support any key that appear. I am not sure how to do it when we have bunch of implemented keys.

teclator: Just expected only the misc commands for using the `MiscRecord` and maybe a `GenericRecord` for the rest not defined yet.

test/ntp_client_test.rb
```diff
@@ -89,6 +90,10 @@
         it "reads sync intervall" do
           expect(subject.sync_interval).to eq 15
         end
+
+        it "do not crash during write" do
```
teclator: does

src/lib/cfa/ntp_conf.rb
```diff
@@ -239,8 +243,22 @@ def self.new_from_augeas(augeas_entry)
...
-        record_class = "::CFA::NtpConf::#{entry_type.capitalize}Record"
-        Kernel.const_get(record_class)
+        record_class = "#{entry_type.capitalize}Record"
+        if CFA::NtpConf.constants.include?(record_class.to_sym)
+          CFA::NtpConf.const_get(record_class.to_sym)
+        else
+          # it is not a predefined record type, so use a generic misc record
+          res = Class.new(MiscRecord)
+          # and specify there its augeas type
+          res.define_method(:augeas_type) { entry_type + "[]" }
```
teclator: defined_singleton_method ?

jreidinger: no, it is not singleton method, it is instance method. But maybe something that can be changed and have it as class method and not instance one.

test/ntp_client_test.rb
```diff
@@ -89,6 +90,13 @@
         it "reads sync intervall" do
           expect(subject.sync_interval).to eq 15
         end
+
+        it "following write succeed" do
```
teclator: NP: it is not clear, at least for me.

teclator:
Just for record purposes, I have tested it manually cloning the current config and writing it back with the expected results.

PR #111, opened 2018-04-13
Master bnc 1072345

schubi2:


src/clients/ntp-client_proposal.rb
```diff
@@ -8,6 +8,8 @@ module Yast
   class NtpClientProposalClient < Client
     include Yast::Logger
 
+    REQUIRED_PACKAGE ||= "chrony".freeze
```
jreidinger: this is wrong as it means it is needed only when client is redefined. It should be solved more by this way - https://github.com/yast/yast-installation/blob/master/src/lib/installation/clients/copy_files_finish.rb#L349

src/lib/y2ntp_client/client/finish.rb
```diff
@@ -4,13 +4,16 @@
...
 Yast.import "NtpClient"
+Yast.import "Pkg"
 
 module Y2NtpClient
   module Client
     # Client to write ntp configuration at the end of installation
     class Finish < Installation::FinishClient
       include Yast::I18n
 
+      REQUIRED_PACKAGE ||= "chrony".freeze
```
jreidinger: same here. And also duplicite. What about adding this constant to NtpClient? It is there as variable, but it can be constant in one place.

src/clients/ntp-client_proposal.rb
```diff
@@ -348,20 +349,18 @@ def Write(param)
...
-      required_package = "chrony"
-
       # In 1st stage, schedule packages for installation
       if Stage.initial
         Yast.import "Packages"
-        Packages.addAdditionalPackage(required_package)
+        Packages.addAdditionalPackage(NtpClientClass::REQUIRED_PACKAGE)
       # Otherwise, prompt user for confirming pkg installation
-      elsif !PackageSystem.CheckAndInstallPackages([required_package])
+      elsif !PackageSystem.CheckAndInstallPackages([NtpClientClass::REQUIRED_PACKAGE])
```
kobliha: Schubi, PLS, see https://github.com/yast/yast-packager/blob/34e7cfcc5a11a038fdbaae23d7f950d385b6bbfb/src/modules/Packages.rb#L922

"OBSOLETE! Please, use PackagesProposal::AddResolvables() instead"

lslezak: This `PackageSystem.CheckAndInstallPackages` is OK the comment belongs to the changed line above.

PR #113, opened 2018-08-20
Fix disabling cron

jreidinger:


jreidinger:
@teclator problem is that this client is not covered at all and not trivial to write.

PR #116, opened 2018-09-07
Save the chronyd service according to user preference

dgdavid:
Related to https://bugzilla.suse.com/show_bug.cgi?id=1075039

---

### Issue

Despite user unchecks "Run NTP as daemon", the chronyd service keeps active and enabled.

### Why?

It seems that the timezone dialog (_yast-country_) is missing the `run_service` param at the time to call the `NtpClientProposalClient#Write` (_yast-ntp-client_). See [yast-country/timezone/src/include/timezone/dialogs.rb:L980](https://github.com/yast/yast-country/blob/a575e590d60657b028bc2003964ec158285618d1/timezone/src/include/timezone/dialogs.rb#L980).

In consequence, [since the default value for that param is `true`](https://github.com/yast/yast-ntp-client/blob/6d99c3d88a49f5f83e3053c215259e479b6db2c2/src/clients/ntp-client_proposal.rb#L345), the service is being started and enabled always, without matter what the user has chosen.

### Implemented solution

Since _NtpClient.run\_service_ has stored the correct value according to the user preferences, this value is now used as default value for `run_service` in the `NtpClientProposalClient#Write`.

### DIFF
--- a/test/ntp_client_proposal_test.rb
+++ b/test/ntp_client_proposal_test.rb
@@ -1,0 +1,169 @@
+require_relative "test_helper"
+
+require_relative "../src/clients/ntp-client_proposal.rb"
+
+Yast.import "Packages"
+
+describe Yast::NtpClientProposalClient do
+  subject do
+    client = described_class.new
+    client.main
+    client
+  end
+
+  describe "#Write" do
+    let(:ntp_server) { "fake.pool.ntp.org" }
+    let(:write_only) { false }
+    let(:ntpdate_only) { true }
+    let(:params) do
+      {
+        "server"       => ntp_server,
+        "write_only"   => write_only,
+        "ntpdate_only" => ntpdate_only
+      }
+    end
+    let(:ntp_client) { Yast::NtpClient }
+    let(:initial_stage) { false }
+    let(:network_running) { false }
+
+    before do
+      allow(subject).to receive(:WriteNtpSettings)
+
+      allow(Yast::Stage).to receive(:initial).and_return(initial_stage)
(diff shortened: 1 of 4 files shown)


test/ntp_client_proposal_test.rb
```diff
@@ -0,0 +1,169 @@
+require_relative "test_helper"
+
+require_relative "../src/clients/ntp-client_proposal.rb"
```
jreidinger: this is a bit problematic, as it execute client program. In this case it just write error to log, but it can change in future ( e.g. many client raise exception if parameter missing or start UI). That is reason why code in src/clients have also zero code coverage.

dgdavid: Ops!

Sorry, I forgot that. Probably you already told me in a previous PR.

So, shall I to discard this test? Or mock the call to `log.error` in this case and keep the test?

What do you prefer or think is better?

Thank you.

jreidinger: keep it as it is for now

lslezak:
:heavy_check_mark: Public Jenkins [job #5](http://ci.opensuse.org/job/yast-yast-ntp-client-master/5/) successfully finished
:heavy_check_mark: Created OBS submit [request #634715](https://build.opensuse.org/request/show/634715)

lslezak:
:heavy_check_mark: Internal Jenkins [job #5](https://ci.suse.de/job/yast-yast-ntp-client-master/5/) successfully finished
:heavy_check_mark: Created IBS submit [request #171738](https://build.suse.de/request/show/171738)

PR #122, opened 2018-11-08
Merge SLE-12-SP3 into SLE-12-SP4

teclator:
It merges #121

### DIFF
--- a/src/clients/ntp-client_proposal.rb
+++ b/src/clients/ntp-client_proposal.rb
@@ -1,13 +1,13 @@
 # encoding: utf-8
 
-# File:	clients/ntp-client_proposal.ycp
-# Summary:	Installation client for ntp configuration
-# Author:	Bubli <kmachalkova@suse.cz>
-#
-# This is used as the general interface between yast2-country
-# (time,timezone) and yast2-ntp-client.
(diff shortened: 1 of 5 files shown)


src/clients/ntp-client_proposal.rb
```diff
@@ -339,76 +339,47 @@ def WriteNtpSettings(ntp_servers, ntp_server, run_service)
...
+    # @return [Symbol] :invalid_hostname, when a not valid ntp_server is given
+    #                  :ntpdate_failed, when the ntp sychronization fails
+    #                  :success, when settings (and sync if proceed) were performed successfully
+    def Write(params)
+      log.info "ntp client proposal Write with #{params.inspect}"
 
-      WriteNtpSettings(ntp_servers, ntp_server, run_service)
-      return :success if param["write_only"]
+      # clean params
+      params.delete_if { |_, v| v.nil? }
```
jreidinger: which ruby version has SP4? Still too old?

teclator: I was asking me the same question :), but as it only needed this change for SP4 I let it as it is and planned to modify for SLE-15-GA

PR #124, opened 2018-11-08
Merge SLE-15-GA into master

teclator:
It merges #123

### DIFF
--- a/src/clients/ntp-client_proposal.rb
+++ b/src/clients/ntp-client_proposal.rb
@@ -356,11 +356,11 @@ module Yast
 
       return :invalid_hostname unless ValidateSingleServer(ntp_server)
 
-      WriteNtpSettings(ntp_servers, ntp_server, run_service)
+      add_or_install_required_package unless params["write_only"]
+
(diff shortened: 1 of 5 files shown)


src/clients/ntp-client_proposal.rb
```diff
@@ -356,11 +356,11 @@ def Write(params)
...
       return :invalid_hostname unless ValidateSingleServer(ntp_server)
 
-      WriteNtpSettings(ntp_servers, ntp_server, run_service)
+      add_or_install_required_package unless params["write_only"]
 
-      return :success if params["write_only"]
+      WriteNtpSettings(ntp_servers, ntp_server, run_service) unless params["ntpdate_only"]
 
-      add_or_install_required_package
+      return :success if params["write_only"]
```
lslezak: Could we simplify it to 

```ruby
return :success if params["write_only"]

add_or_install_required_package
WriteNtpSettings(ntp_servers, ntp_server, run_service)
```

Or did I overlook something?

teclator: In that case, when called for only writing, it will return :success without writing, or I'm now overlooking something?

PR #133, opened 2019-03-08
Bump version & changelog

teclator:
The popup was added in https://github.com/yast/yast-yast2/pull/831

But was still not available in the MU.

See: https://build.suse.de/package/live_build_log/SUSE:Maintenance:9763/yast2-ntp-client.SUSE_SLE-15_Update/SUSE_SLE-15_Update/x86_64


package/yast2-ntp-client.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Fri Mar  8 14:39:23 UTC 2019 - knut.anderssen@suse.com
+
+- Added cwm/popup dependency (bnc#893065, bsc#1039985)
```
jreidinger: please use in both case same shortcut. or if you want emphasize suse internal as bsc and opensuse bug as boo I am fine. but bnc is now dead and should not be used anymore.

PR #137, opened 2019-06-13
Enable/disable chronyd service: Enable/disable chrony-wait

schubi2:


src/modules/NtpClient.rb
```diff
@@ -594,6 +595,7 @@ def dhcp_ntp_servers
     publish variable: :synchronize_time, type: "boolean"
     publish variable: :sync_interval, type: "integer"
     publish variable: :service_name, type: "string"
+    publish variable: :wait_service_name, type: "string"
```
jreidinger: do we need to publish this variable to component system? Is it used outside?

src/modules/NtpClient.rb
```diff
@@ -92,8 +92,9 @@ def main
       # The interval of synchronization in minutes.
       @sync_interval = DEFAULT_SYNC_INTERVAL
 
-      # Service name of the NTP daemon
+      # Service names of the NTP daemon
       @service_name = "chronyd"
+      @wait_service_name = "chrony-wait"
```
jreidinger: can you please add here comment why it is needed, ideally with reference to bugzilla.

test/ntp_client_test.rb
```diff
@@ -283,10 +283,37 @@
...
+    context "services will be started" do
+      before do
+        subject.run_service = true
+        subject.write_only = false
+      end
 
-      subject.Write
+      it "enables and restarts services" do
+        allow(subject).to receive(:check_service).and_call_original
+        expect(Yast::Service).to receive(:Enable).with(subject.service_name).and_return(true)
```
jreidinger: I think better it to use explicit string. So this way accessor won't be needed.

src/modules/NtpClient.rb
```diff
@@ -795,24 +797,37 @@ def write_and_update_policy
...
+        if !Service.Enable(@service_name)
+          Report.Error(Message.CannotAdjustService(@service_name))
+        elsif !Service.Enable(@wait_service_name)
+          Report.Error(Message.CannotAdjustService(@wait_service_name))
+        end
+      elsif !Service.Disable(@service_name)
+        Report.Error(Message.CannotAdjustService(@service_name))
+      elsif !Service.Disable(@wait_service_name)
+        Report.Error(Message.CannotAdjustService(@wait_service_name))
+      end
```
mchf: this deserves splitting into own method.

src/modules/NtpClient.rb
```diff
@@ -795,24 +797,37 @@ def write_and_update_policy
...
-    # * When enabling, it tries to restart the service unless it's in write
+    # * When disabling, it also stops the services.
+    # * When enabling, it tries to restart the services unless it's in write
     #   only mode.
     def check_service
-      adjusted = @run_service ? Service.Enable(@service_name) : Service.Disable(@service_name)
-
-      # error report
-      Report.Error(Message.CannotAdjustService("NTP")) unless adjusted
+      if @run_service
```
mchf: this method contains two separate workflows enclosed in two ```if @run_service``` branches. Moreover first of them do not contain else part. It can be rewritten to avoid one level of nesting e.g. like
```
def check_service
  if !@run_service
    Service.Stop(@service_name)
    Service.Stop(@wait_service_name)
    return
  end
....
end
````
which seems to be more readable for me.

mchf:
Apart of check_service name which i see a bit misleading it lgtm. So, lets wait in @jreidinger comments.

PR #139, opened 2019-07-05
Fix "ntp-client" return value for SLE 15 SP1

imobachgs:
Bring #139 to `SLE-15-SP1`.

### DIFF
--- a/src/lib/y2ntp_client/dialog/main.rb
+++ b/src/lib/y2ntp_client/dialog/main.rb
@@ -5,7 +5,7 @@ require "y2ntp_client/widgets/main_widgets"
 
 Yast.import "Label"
 Yast.import "NtpClient"
-Yast.import "Stage"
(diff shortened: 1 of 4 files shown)


src/lib/y2ntp_client/dialog/main.rb
```diff
@@ -70,13 +71,23 @@ def hardware_clock_widgets
...
       def next_button
-        # FIXME: it probably cannot run in initial stage, so not needed
-        Yast::Stage.initial ? Yast::Label.AcceptButton : Yast::Label.OKButton
+        return Yast::Label.OKButton unless installation?
+        nil
+      end
+
+      # Determines whether running in installation mode
+      #
+      # We do not use Stage.initial because of firstboot, which which runs in 'installation' mode
```
dgdavid: NP: `which` is duplicated.

PR #144, opened 2019-08-19
AutoYaST: sanitize comments when importing

joseivanlopez:
### Problem

When an imported AutoYaST profile specifies an NTP entry where the comment contains empty lines, the Augeas parser fails when trying to save the `/etc/ntp.conf` file.

```
      <peer>
        <address>1.opensuse.pool.ntp.org</address>
        <comment> # a comment with spaces

          # and blank lines

        </comment>
        <options>iburst</options>
        <type>server</type>
      </peer>
```

*  https://bugzilla.suse.com/show_bug.cgi?id=1142026

### Solution

This was already fixed by https://github.com/yast/yast-ntp-client/pull/143. IMHO, the problem with that solution is that we are delegating the responsibility of fixing the imported record to the method `NtpConf#save`.

This new solution moves such responsibility to the `sanitizer` method that already exists. This sanitizer is called during the AutoYaST importation. 

### Tests

* Added unit tests
* Manually tested

### DIFF
--- a/src/modules/NtpClient.rb
+++ b/src/modules/NtpClient.rb
@@ -1018,22 +1018,63 @@ module Yast
 
   private
 
-    # Remove blank spaces in values
+    # Sanitizes values of a record (mainly, by removing blank spaces)
     #
-    # @note to avoid augeas parsing errors, comments should be sanitized by
-    #   removing blank spaces at the beginning and adding line break. Further
-    #   sanitation is done in NtpConf.save.
+    # @param record [Hash]
+    # @return [Hash]
     def sanitize_record(record)
       sanitized = record.dup
+
       sanitized.each do |key, value|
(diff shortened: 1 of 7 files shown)


src/modules/NtpClient.rb
```diff
@@ -1018,24 +1018,65 @@ def AutoPackages
...
+    # @return [String] sanitized comment
+    def sanitize_comment(comment)
+      lines = comment.split("\n")
+
+      sanitized_lines = lines.map { |l| sanitize_comment_line(l) }
+
+      sanitized_lines.compact.join
+    end
+
+    # Sanitizes a a comment line by removing blank spaces at the beginning and ensuring a line break
```
dgdavid: NP: `s/a a/a`

PR #150, opened 2019-10-03
Fix calling autoclient

jreidinger:
https://bugzilla.suse.com/show_bug.cgi?id=1152159


jreidinger:
already fixed by https://github.com/yast/yast-yast2/pull/964

PR #151, opened 2019-10-14
Systemd timer

jreidinger:


jreidinger:
failing as rubocop does not know old heredoc syntax, so this pr depends on https://github.com/yast/yast-ntp-client/pull/152

src/modules/NtpClient.rb
```diff
@@ -35,8 +35,22 @@ class NtpClientClass < Module
...
-    CRON_FILE = "/etc/cron.d/suse-ntp_synchronize".freeze
+    TIMER_FILE = "yast-timesync.timer"
+    # The file name of systemd timer for the synchronization.
+    TIMER_PATH = "/etc/systemd/system/#{TIMER_FILE}".freeze
+
+    # Timer content which should be passed to format and seconds added.
+    # @example usage
+    # format(TIMER_CONTENT, timeout: 15)
+    TIMER_CONTENT = <<~TIMER
+      [Timer]
```
lslezak: I usually do not like heredoc strings as they make editing less convenient, e.g. the editors usually do not support syntax highlighting, syntax checks, etc...

I'd prefer to put it into a separate template file, e.g. `yast-timesync.timer.erb` which would contain the ERB template. You can render it like [here](https://github.com/yast/yast-registration/blob/9d6e06ead9f5fb784ab38bb34b9995ca72cd946d/src/lib/registration/helpers.rb#L165) in yast2-registration.

This would also avoid the duplication of the content in the *.spec file, you could simply call

    /usr/bin/erb timeout=<timeout> /.../yast-timesync.timer.erb > /etc/systemd/system/yast-timesync.timer

instead of the copy&paste string in the *.spec file. This would ensure that the content is the same, we do not need to fix two places. That's fragile.

src/modules/NtpClient.rb
```diff
@@ -281,15 +295,15 @@ def ProcessNtpConf
...
 
-      sync_interval_entry = cron_entry.fetch("minute", "*/#{DEFAULT_SYNC_INTERVAL}")
-      log.info("MINUTE #{sync_interval_entry}")
+      timer_content = File.read(TIMER_PATH)
+      log.info("NTP Synchronization timer entry: #{timer_content}")
+      @synchronize_time = SCR.Execute(path(".target.bash"),
+        "/bin/systemctl is-active #{TIMER_FILE}").zero?
 
-      @sync_interval = sync_interval_entry.tr("^[0-9]", "").to_i
+      interval = timer_content[/^\s*OnUnitActiveSec=(\d+)m/, 1]
```
lslezak: I do not know the exact syntax, are the spaces around `=` allowed? Maybe the regexp should be less strict about the expected format.

src/modules/NtpClient.rb
```diff
@@ -281,15 +295,15 @@ def ProcessNtpConf
...
-      cron_entry = (crontab || []).fetch(0, {}).fetch("events", []).fetch(0, {})
-      @synchronize_time = cron_entry["active"] == "1"
+      return false unless File.exist?(TIMER_PATH)
 
-      sync_interval_entry = cron_entry.fetch("minute", "*/#{DEFAULT_SYNC_INTERVAL}")
-      log.info("MINUTE #{sync_interval_entry}")
+      timer_content = File.read(TIMER_PATH)
+      log.info("NTP Synchronization timer entry: #{timer_content}")
+      @synchronize_time = SCR.Execute(path(".target.bash"),
+        "/bin/systemctl is-active #{TIMER_FILE}").zero?
```
lslezak: Maybe we could use some existing functionality for this? (`Service` module?)

test/ntp_client_test.rb
```diff
@@ -437,23 +437,24 @@
...
   describe "#ReadSynchronization" do
-    let(:cron_job_file) { "/etc/cron.d/suse-ntp_synchronize" }
-    let(:cron_entry) { [] }
+    let(:systemd_timer_file) { "/etc/systemd/system/yast-timesync.timer" }
+    let(:timer_content) { "" }
 
     before do
-      allow(Yast::SCR).to receive(:Read)
-        .with(Yast::Path.new(".cron"), cron_job_file, "").and_return(cron_entry)
+      allow(::File).to receive(:exists?).and_return(true)
```
lslezak: `File.exists?` is deprecated and should not be used, use `File.exist?` instead.

test/ntp_client_test.rb
```diff
@@ -437,23 +437,24 @@
...
+      expect(::File).to receive(:read).and_return(timer_content)
 
       subject.ReadSynchronization
     end
 
-    context "when cron file does not exist" do
-      let(:cron_entry) { nil }
+    context "when systemd timer file does not exist" do
+      before do
+        allow(::File).to receive(:exists?).and_return(false)
```
lslezak: The same here.

src/modules/NtpClient.rb
```diff
@@ -839,19 +853,35 @@ def check_service
...
+    def update_timer_settings
       if @synchronize_time
         SCR.Write(
           path(".target.string"),
-          CRON_FILE,
-          "-*/#{@sync_interval} * * * * root /usr/sbin/chronyd -q &>/dev/null\n"
+          TIMER_PATH,
+          format(TIMER_CONTENT, timeout: @sync_interval)
+        )
+        SCR.Execute(
```
lslezak: Again, no support in YaST for this? I'd avoid directly calling `systemctl` if possible.

jreidinger: well, we have Systemctl class, but in general now we only support service, sockets and targets, not timers.

test/ntp_client_test.rb
jreidinger: rubocop fix it itself.

package/yast2-ntp-client.spec
```diff
@@ -69,14 +71,43 @@ This package contains the YaST2 component for NTP client configuration.
 %install
 %yast_install
 %yast_metainfo
+install -D -m 644 %{S:1} %{buildroot}%{_unitdir}/yast-timesync.service
 
 %post
+%service_add_post yast-timesync.service
+
 # upgrade old name and convert it to chrony (bsc#1079122)
 if [ -f /etc/cron.d/novell.ntp-synchronize ]; then
```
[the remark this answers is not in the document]
jreidinger: well, I prefer to have to straightforward. Doing everything in one step is more error prone and also harder to test for me. So I keep working step and add one more step that I test manually if it works.

package/yast2-ntp-client.spec
```diff
@@ -69,14 +71,43 @@ This package contains the YaST2 component for NTP client configuration.
...
+  cat <<EOT > /etc/systemd/system/yast-timesync.timer
+[Timer]
+# first sync after boot
+OnBootSec=1min
+OnUnitActiveSec=$(grep -o '[[:digit:]]\+' /etc/cron.d/suse-ntp_synchronize)min
+
+[Install]
+WantedBy=timers.target
+EOT
+  /bin/systemctl enable yast-timesync.timer
```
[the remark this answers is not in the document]
jreidinger: I think not, as it is discouraged in general in rpm spec and using preset instead https://en.opensuse.org/openSUSE:Systemd_packaging_guidelines#Enabling_systemd_unit_files

but we need it for convert of old way.

src/modules/NtpClient.rb
lslezak: Ah, I see, OK.

package/yast2-ntp-client.spec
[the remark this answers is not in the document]
lslezak: OK

src/lib/y2ntp_client/yast-timesync.timer.erb
```diff
@@ -0,0 +1,7 @@
+[Timer]
```
lslezak: I'm just not sure if `lib` subdir is the best place for this, it's a data file, you cannot `require` it in Ruby.

In registration I use [`data/registration`](https://github.com/yast/yast-registration/tree/master/src/data/registration) subdirectory (which is installed into `/usr/share/YaST2/data/registration/`).

package/yast2-ntp-client.spec
```diff
@@ -71,12 +72,32 @@ This package contains the YaST2 component for NTP client configuration.
...
 # upgrade old name and convert it to chrony (bsc#1079122)
 if [ -f /etc/cron.d/novell.ntp-synchronize ]; then
   mv /etc/cron.d/novell.ntp-synchronize /etc/cron.d/suse-ntp_synchronize
   sed -i 's:\* \* \* \* root .*:* * * * root /usr/sbin/chronyd -q \&>/dev/null:' /etc/cron.d/suse-ntp_synchronize
 fi
 
+# and now update cron to systemd timer. We need to support upgrade from SLE12 and also SLE15 SP1.
+# jsc#SLE-9113
+if [ -f /etc/cron.d/suse-ntp_synchronize ]; then
+  /usr/bin/erb timeout=$(grep -o '[[:digit:]]\+' /etc/cron.d/suse-ntp_synchronize) /usr/share/YaST2/lib/y2ntp_client/yast-timesync.timer.erb > /etc/systemd/system/yast-timesync.timer
```
lslezak: You need to update also this ERB file location.

PR #156, opened 2020-03-03
Supporting more than one ntp server in the time setting module.

schubi2:
**Problem**
The user can select **one** server in the time setting module and can switch to the ntp client module in order to add/update additional one. So the settings/views in both screen are not compatible together.
**Solution**
If more than one ntp server has configured in the ntp client they will be only shown in the time setting screen
ntp (1)
**Testing**
Manuel testing. Adapting test cases.

### DIFF
--- a/src/clients/ntp-client_proposal.rb
+++ b/src/clients/ntp-client_proposal.rb
@@ -384,7 +384,7 @@ module Yast
 
       return :success if params["write_only"]
 
-      # Only if network is running and user has selected an ntp server try to synchronize
+      # Only if network is running try to synchronize
       # the ntp server.
       if NetworkService.isNetworkRunning && !Service.Active(NtpClient.service_name)
         if !select_ntp_server && ntp_server.empty?
@@ -411,7 +411,7 @@ module Yast
             UI.QueryWidget(Id(:ntp_address), :Value)
           )
         elsif rv == :next && !Stage.initial
-          # Update UI for the changes ntp servers
+          # Updating UI for the changed ntp servers
           ui_init(Id(:rp), false)
           # show the 'save' status after configuration
           UI.ChangeWidget(Id(:ntp_save), :Value, GetNTPEnabled())
--- a/src/clients/ntp-client_proposal.rb
+++ b/src/clients/ntp-client_proposal.rb
@@ -221,7 +221,7 @@ module Yast
         if NtpClient.GetUsedNtpServers.size > 3
           # TRANSLATOR %{count} number of additional servers
           text << format(_("... (%{count} more servers)"),
-             count: (NtpClient.GetUsedNtpServers.size - counter))
+            count: (NtpClient.GetUsedNtpServers.size - counter))
         end
         ntp_server_widget = Label(text)
       end
--- a/src/clients/ntp-client_proposal.rb
+++ b/src/clients/ntp-client_proposal.rb
@@ -555,6 +555,7 @@ module Yast
       # module which is not defined in the combo box. In that case we do not offer
       # a selection. The user should go back to ntp-client to change it.
       if NtpClient.GetUsedNtpServers.size == 1
+        puts "llll #{fallback_ntp_items}"
         ret = fallback_ntp_items.any? do |item|
           item.params[1] == NtpClient.GetUsedNtpServers.first
         end
--- a/src/clients/ntp-client_proposal.rb
+++ b/src/clients/ntp-client_proposal.rb
@@ -128,7 +128,7 @@ module Yast
     end
 
     def ui_enable_disable_widgets(enabled)
-      UI.ChangeWidget(Id(:ntp_address), :Enabled, enabled)
+      UI.ChangeWidget(Id(:ntp_address), :Enabled, enabled) if select_ntp_server
       UI.ChangeWidget(Id(:run_service), :Enabled, enabled)
       # FIXME: With chronyd, we cannot synchronize if the service is already
       # running, we could force a makestep in this case, but then the button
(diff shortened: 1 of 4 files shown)


src/clients/ntp-client_proposal.rb
```diff
@@ -289,16 +302,16 @@ def ui_init(rp, first_time)
...
+        ntp_server = Convert.to_string(UI.QueryWidget(Id(:ntp_address), :Value))
+        return :invalid_hostname unless ValidateSingleServer(ntp_server)
+
         NtpClient.ntp_conf.clear_pools
         NtpClient.ntp_conf.add_pool(ntp_server)
-        retval = Convert.to_boolean(WFM.CallFunction("ntp-client"))
-        ret = :next if retval
-        MakeProposal()
       end
+      ret = :next if WFM.CallFunction("ntp-client")
```
jreidinger: will it work when you just pick just one entry? Comment for method really missing.

src/clients/ntp-client_proposal.rb
```diff
@@ -514,16 +532,34 @@ def dhcp_ntp_items
...
+        @cached_fallback_ntp_items = timezone_ntp_items
       end
+      @cached_fallback_ntp_items
+    end
 
-      log.info("Proposing NTP server list provided by DHCP")
-      dhcp_items
+    # Checking if the user can select one ntp server from the list
+    # of proposed servers.
+    # It does not make sense if there are more than one ntp server has been
```
jreidinger: double verb, probably you want something like `It does not make sense if there are more than one ntp server defined`

src/clients/ntp-client_proposal.rb
```diff
@@ -514,16 +532,34 @@ def dhcp_ntp_items
...
+      ret = NtpClient.GetUsedNtpServers.nil? || NtpClient.GetUsedNtpServers.empty?
+      # It could be that the user has defined an own ntp server in the ntp-client
+      # module which is not defined in the combo box. In that case we do not offer
+      # a selection. The user should go back to ntp-client to change it.
+      if NtpClient.GetUsedNtpServers.size == 1
+        ret = fallback_ntp_items.any? do |item|
+          item.params[1] == NtpClient.GetUsedNtpServers.first
+        end
+      end
+      ret
```
jreidinger: well, this part of code is already big and not tested at all. And this change again increase it. I think we should consider moving this part of code somewhere where it can be tested and also some refactoring is welcome.

jreidinger: ah, it is tested, so it is a bit better. Still I think it would be nice to not increase this file and instead e.g. create own dialog class that is used from this client.

src/clients/ntp-client_proposal.rb
```diff
@@ -436,8 +458,11 @@ def ui_try_save
 
       rv = Write(argmap)
 
-      server = Convert.to_string(UI.QueryWidget(Id(:ntp_address), :Value))
-
+      server = if select_ntp_server
```
jreidinger: will it work below? especially https://github.com/yast/yast-ntp-client/pull/156/files#diff-637cd929308624ec8989c144f1d7b7c6R484 where it will write this first server? Also I am not sure if it is proper behavior if you have set of servers that it query only the first one.

src/clients/ntp-client_proposal.rb
```diff
@@ -370,8 +385,13 @@ def Write(params)
 
       return :success if params["write_only"]
 
-      # Only if network is running try to synchronize the ntp server
+      # Only if network is running try to synchronize
+      # the ntp server.
       if NetworkService.isNetworkRunning && !Service.Active(NtpClient.service_name)
+        if !select_ntp_server && ntp_server.empty?
+          # Trying first server in the list
+          ntp_server = NtpClient.GetUsedNtpServers.first unless NtpClient.GetUsedNtpServers.nil?
```
jreidinger: and same here with list of servers. If you have multiple servers and it sync only against one and that one failing, it fails whole, which is strange and not how works list of ntp servers. Also time will be less precise then with multiple sources.

src/clients/ntp-client_proposal.rb
```diff
@@ -208,19 +202,38 @@ def MakeProposal
...
+        # Only show all ntp servers
+        text = _("Synchronization Servers:\n").dup
+        counter = NtpClient.GetUsedNtpServers.size > 3 ? 3 : NtpClient.GetUsedNtpServers.size
+        counter.times do |i|
+          text << NtpClient.GetUsedNtpServers[i]
+          text << "\n"
+        end
+        if NtpClient.GetUsedNtpServers.size > 3
+          # TRANSLATOR %1 number of additional servers
+          text << Builtins.sformat(_("... (%1 more servers)"),
```
jreidinger: for new code I would not use Builtns.

src/clients/ntp-client_proposal.rb
schubi2: Yes, it is working. I have have added comments.

src/clients/ntp-client_proposal.rb
schubi2: Well, espl. for that part I have added testcases in order to be sure.

src/clients/ntp-client_proposal.rb
schubi2: fixed.

src/clients/ntp-client_proposal.rb
schubi2: Yes, I have updated it. Now each ntp server will be tried until one has succeeded.

src/clients/ntp-client_proposal.rb
schubi2: I have removed it.

src/clients/ntp-client_proposal.rb
schubi2: Fixed.

src/clients/ntp-client_proposal.rb
schubi2: Well this is an L3. I do not want a refactoring. 
To be honest this file has to be refactored completely. This already begins with the complete wrong file name and ends in the code. But first of all we have to decide if the connection of time settings and ntp-client configuration makes sense at all. At first we should decide how the UI should be before starting a refactoring.

PR #158, opened 2020-03-18
 Do not show checkboxes for saving configuration and starting the deam…

schubi2:
…on in an installaed system

This is a follow up of:
https://github.com/yast/yast-ntp-client/pull/156

### DIFF
--- a/src/clients/ntp-client_proposal.rb
+++ b/src/clients/ntp-client_proposal.rb
@@ -89,56 +89,56 @@ module Yast
     end
 
     def ui_help_text
-      # help text
-      tmp = _(
-        "<p>Press <b>Synchronize Now</b>, to get your system time set correctly " \
-        "using the selected NTP server. If you want to make use of NTP permanently, " \
-        "enable the <b>Save NTP Configuration</b> option</p>"
-      )
-
-      tmp = Ops.add(
-        tmp,
-        _(
-          "<p>Enabling <b>Run NTP as daemon</b> option, the NTP service will be " \
-          "started as daemon. Otherwise the system time will be synchronized periodically. " \
-          "The default interval is 15 min. You can change it after installation " \
-          "with the <b>yast2 ntp-client module</b>.</p>"
+      if Stage.initial
+        # help text
+        tmp = _(
+          "<p>Press <b>Synchronize Now</b>, to get your system time set correctly " \
+          "using the selected NTP server. If you want to make use of NTP permanently, " \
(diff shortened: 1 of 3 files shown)


src/clients/ntp-client_proposal.rb
```diff
@@ -89,56 +89,56 @@ def main
...
+            "The default interval is 15 min. You can change it after installation " \
+            "with the <b>yast2 ntp-client module</b>.</p>"
+          )
+        )
+        # help text, cont.
+        tmp = Ops.add(
+          tmp,
+          _(
+            "<p>Synchronization with the NTP server can be done only when " \
+            "the network is configured.</p>"
```
jreidinger: when you touching it, can you remove that Ops and use simple `_() + _() + _()` ?

src/clients/ntp-client_proposal.rb
```diff
@@ -89,56 +89,56 @@ def main
...
       UI.ChangeWidget(Id(:ntp_address), :Enabled, enabled) if select_ntp_server
-      UI.ChangeWidget(Id(:run_service), :Enabled, enabled)
-      # FIXME: With chronyd, we cannot synchronize if the service is already
-      # running, we could force a makestep in this case, but then the button
-      # should be reworded and maybe the user should confirm it (bsc#1087048)
-      if !NetworkService.isNetworkRunning || Service.Active(NtpClient.service_name)
-        UI.ChangeWidget(Id(:ntp_now), :Enabled, false)
-      else
-        UI.ChangeWidget(Id(:ntp_now), :Enabled, enabled)
+      if Stage.initial
```
jreidinger: BTW when seeing all this conditions I start thinking that it would be nicer to have own class for runtime and installation. I know it is not possible for maintenance branch, but in general it can make code much cleaner as now runtime and installation differs a lot.

schubi2: Yes, we have already opened a new card for refactoring the complete code.

src/clients/ntp-client_proposal.rb
```diff
@@ -89,56 +89,46 @@ def main
...
 
-      # help text, cont.
-      tmp = Ops.add(
-        tmp,
-        _(
-          "<p>Synchronization with the NTP server can be done only when " \
-          "the network is configured.</p>"
-        )
-      )
       tmp
```
jreidinger: and now when I see it, whole tmp can be removed as it is not needed and can be simple method like
```
def a
  if Stage.initial
    _() + _() + _()
  else
    _()
  end
end

PR #173, opened 2022-05-02
Updated ntp sources handling

mchf:
## Problem

For several reasons ntp sources handling is not much user friendly.

For example in OpenSUSE installation the dialog looks this way
ntp-opensuse

There is no possibility to change listed ntp sources (add / remove).
The code doesn't distinguish between ntp server and pool.
What is less visible is that sources are written into "wrong" config file.
Behavior in SLE is not exactly the same but similar enough. 

What are major caveats?
1) the dialog shown in the screenshot is not implemented in one module but is spread in two modules (y2-country and y2-ntp-client)
2) similarly to dialog creation, the handling is done in similar way. So, there is bunch of ```ui_``` prefixed methods in y2-ntp-client's proposal client. Whenever is some operation done in the ntp part of dialog then it is proposed by client's api. It means that timezone dialog's loop handler checks if the event comes from ntp part of dialog, then it calls an ntp-client handler via proposal client's api and receives the response in the same way. It has some drawbacks regarding handling current dialog state.

## Solution

Purpose of this PR was to keep current functionality and open possibilities to implement new behavior. Mainly support to really add / remove ntp sources and distinguish them to pool / servers.

So, as first step the dialog was redesigned to
ntp_patched_4

It allows to add / remove ntp sources and perform real configuration, not only using what installation sets for the user with no possibility to change anything.

It shows knowledge of difference between ntp server / pool.
List of  ntp sources offered by default in the combobox was taken from SLE. User is allowed to write own ntp source.

Ntp sources are also read from the control file and from dhcp. Dhcp sources are not proposed by default, but available for selection in combobox. One of control file sources is randomly picked and proposed. See also previous [yast2-country](https://github.com/yast/yast-country/pull/282) PR for some details.

Pools / servers are written into ```/etc/chrony.d/pool.conf```. The pool.conf is provided by chrony-pool-* package(s) (product specific). We have to overwrite it otherwise yast's setup could be corrupted. The pool.conf is deployed by a package by design for needs of image based (no yast) installations.

## Testing

- *Tested manually*

## Screenshots

See above


src/lib/y2ntp_client/widgets/sources_table.rb
```diff
@@ -0,0 +1,119 @@
...
+
+      def header
+        [
+          _("Type"),
+          _("Address")
+        ]
+      end
+
+      def items
+        @sources || []
```
teclator, line 44: It is really needed? we are initializing it to an array already.

teclator:
```suggestion
        sources
```

src/lib/y2ntp_client/widgets/sources_table.rb
```diff
@@ -0,0 +1,119 @@
...
+module Y2NtpClient
+  module Widgets
+    # Table displaying list of defined NTP sources. It displays its type and address.
+    class SourcesTable < CWM::Table
+      # @param sources [Array<String>] array of ntp sources (ip or url)
+      def initialize(sources = [])
+        textdomain "ntp-client"
+
+        # TODO: kind of validarion pre-processing
+        # <id, source-type, source-address>
```
teclator, line 32: We could add an attr_reader describing the variable although by now it looks like the type is not used...

src/lib/y2ntp_client/widgets/sources_table.rb
```diff
@@ -0,0 +1,119 @@
...
+      # Removes one item from table's content
+      #
+      # @param id [any] id of table's item to remove
+      def remove_item(id)
+        updated_items = items.delete_if { |i| i[0] == id }
+        change_items(updated_items)
+      end
+
+      def config
+        Yast::Lan.yast_config
```
teclator, line 67: Looks like useless method.

src/lib/y2ntp_client/widgets/sources_table.rb
```diff
@@ -0,0 +1,119 @@
...
+          _("Type"),
+          _("Address")
+        ]
+      end
+
+      def items
+        @sources || []
+      end
+
+      def addresses
```
teclator, line 47: looks like another useless method

teclator:
Would be nice to see also how it looks like in **ncurses**. Therefore some screenshot would be appreciated

teclator:
There is a type column but there is no information about the source type at all but in general do not look bad although for UX I would ask @dgdavid  ;)

mchf:
This time this works:
1) combo box is pre-filed with values obtained from dhcp, control file, ....
2) some servers are picked by default - e.g. those obtained from dhcp
3) user is allowed to pick whether the source is pool or server
4) sources are written to the target at the end of installation

Major changes:
1) the dialog is re-entrant. Before it behaved differently depending whether entered first time, or something was selected by user before and some other circumstances
2) user is allowed to store  the source as server or pool
3) sources selected by default are as before

ntp_patched

teclator:
> This time this works:
> 
> 1. combo box is pre-filed with values obtained from dhcp, control file, ....
> 2. some servers are picked by default - e.g. those obtained from dhcp
> 3. user is allowed to pick whether the source is pool or server
> 4. sources are written to the target at the end of installation
> 
> Major changes:
> 
> 1. the dialog is re-entrant. Before it behaved differently depending whether entered first time, or something was selected by user before and some other circumstances
> 2. user is allowed to store  the source as server or pool
> 3. sources selected by default are as before
> 
> ntp_patched

¿What about the servers defined in the rpm? I guess we should not provide default configuration based on rpms anymore in order to not collide with the configured ones or at least we should select the **chrony-pool-empty** package in the proposal

src/clients/ntp-client_proposal.rb
```diff
@@ -1,11 +1,18 @@
 require "yast"
 
+require "y2ntp_client/widgets/sources_table"
+
 module Yast
   # This is used as the general interface between yast2-country
   # (time,timezone) and yast2-ntp-client.
   class NtpClientProposalClient < Client
```
teclator, line 8: Taking into account the big changes you are doing maybe it is time to move it to **lib/y2ntp_client** and use the **YaST2** Installation/ProposalClient and start testing it properly.

mchf:
> 
> ¿What about the servers defined in the rpm? I guess we should not provide default configuration based on rpms anymore in order to not collide with the configured ones or at least we should select the **chrony-pool-empty** package in the proposal

Proposals are currently taken only from dhcp and control file. In my POV, any "default" servers provided by an rpm package should not be used as it is somehow against what the proposal in installer does. However, this is for a discussion somewhere else. We're having a bit of chicken egg problem here, however I'd say that we need to have working installer proposal accepting user modifications and then we can ask for some more changes elsewhere.

teclator:
> Proposals are currently taken only from dhcp and control file. In my POV, any "default" servers provided by an rpm package should not be used as it is somehow against what the proposal in installer does. 

I agree, and it was part of the PBI to discuss it first before any implementation.

>However, this is for a discussion somewhere else. 

+1

>We're having a bit of chicken egg problem here, however I'd say that we need to have working installer proposal accepting user modifications and then we can ask for some more changes elsewhere.

Not at all, we need to define behavior and implement accordingly, if we not show the user the current servers then it could be problematic and a mess which is exactly what is currently happening unless the user deselect the default rpm and selects the empty pools one, so this could be part of the current proposal until the product patterns or chrony package recommends are adapted.

BTW, checking the current implementation with a DUD ends with a long list of servers (known ones if I'm not wrong)

Screenshot_testing_2022-11-23_10:04:57

I expected the ones from the control file but that is not the case, is there some change in yast2-country needed by this PR?

src/clients/ntp-client_proposal.rb
```diff
@@ -1,11 +1,18 @@
...
     include Yast::Logger
 
+    @sources_table = nil
```
teclator, line 11: I would omit this initialization and move it to a method...
```suggestion
    def sources_table
      @sources_table_widget ||= Y2NtpClient::Widgets::SourcesTable.new(NtpClient.GetUsedNtpSources)
    end
```

mchf:
Minor fix in default initialization of ntp sources dialog
ntp_patched_2

src/clients/ntp-client_proposal.rb
mchf: sorry but the client is used in a way that trying to do it properly will be another major step which will also include redesigning yast2-country

dgdavid:
Just a minor suggestion, slightly out of the scope of this PR: why not take the opportunity to put _Current time_ and _Current date_ inline?

I mean, when selecting _Manually_ see something like 

```
Date        Time
▒▒▒▒▒▒▒▒▒   ▒▒▒▒▒▒▒▒▒
```

instead of the current

```
   
Time
▒▒▒▒▒▒▒▒▒  

Date     
▒▒▒▒▒▒▒▒▒
```

mchf:
> Just a minor suggestion, slightly out of the scope of this PR: why not take the opportunity to put _Current time_ and _Current date_ inline?
> 
> I mean, when selecting _Manually_ see something like
> 
> ```
> Date        Time
> ▒▒▒▒▒▒▒▒▒   ▒▒▒▒▒▒▒▒▒
> ```
> 
> instead of the current
> 
> ```
>    
> Current Time
> ▒▒▒▒▒▒▒▒▒  
> 
> Current Date     
> ▒▒▒▒▒▒▒▒▒
> ```

it is in different module (yast2-country) see the description how it works to get into fun with this dialog ;-)

src/modules/NtpClient.rb
```diff
@@ -33,7 +33,7 @@ class NtpClientClass < Module
     # @see #http://www.pool.ntp.org/
     RANDOM_POOL_NTP_SERVERS = ["0.pool.ntp.org", "1.pool.ntp.org", "2.pool.ntp.org"].freeze
 
-    NTP_FILE = "/etc/chrony.conf".freeze
+    NTP_FILE = "/etc/chrony.d/pool.conf".freeze
```
teclator, line 36: So, the pool.conf will not contain only pool.conf but also servers and probably other entries there isn't it? Maybe we should have our own config file as we do with sysctl config.

teclator:
Screenshot from 2022-12-21 09-51-10

src/clients/ntp-client_proposal.rb
```diff
@@ -203,13 +203,17 @@ def MakeProposal
       # Once read or proposed any config we consider it as read (bnc#427712)
       NtpClient.config_has_been_read = true
 
-      # do a proposal
-      ntp_sources = fallback_ntp_items
```
teclator: Why are we modifying the defaults? In case no one is provided it should use the defaults ones from the control file isn't it? at least we are using them for the first sync.

teclator: So, we modified it because the sources table was filled by the Country ntp servers but I guess we should use https://github.com/yast/yast-network/blob/9874173c8b9a15135da12621e4f50522cc3ec9dc/src/lib/y2network/ntp_server.rb#L58

mchf: It is some magic which is done across the live of the dialog. 

First initialization is done in yast2-country, where sources from control files are read. However, later then in the ntp module some more logic is added. Like reading sources obtained from dhcp, "default hard coded" geo based servers. 

So, it simply is cumulation of hacks as different features got implemented over the time.

mchf:
> Screenshot from 2022-12-21 09-51-10
> 
> So, rubocop is now happy but change_items does not return a collection ...
> 
> [a18efd1](https://github.com/yast/yast-ntp-client/commit/a18efd11b63bb8a250164e2c99da19ca5e8deb5b)

interesting what one forgotten braces can cause. Should be fixed and what is better is that rubocop is still happy.

src/modules/NtpClient.rb
mchf: either we can split it into "server.conf" and "pool.conf", however it sounds a bit overkill to me ...  or have some neutral "sources.conf" with all together as it is now (but with better name). I'm not good in naming so I didn't spend much time in the area yet and I'm open to name suggestions ;-)

mchf:
I've modified default servers proposal. According to fate#323454  and bsc#1188980 we should propose a suse's pool and do not propose dhcp provided servers (but offer them for selection). So, the proposal looks like this:
ntp_patched_3

src/clients/ntp-client_proposal.rb
mchf: I've modified this piece to make it work according to fate#323454) and bsc#1188980 hopefuly.

src/modules/NtpClient.rb
mchf: one benefit if we write it to pool.conf is that we need to deal with one provided by default by chrony

mchf:
I've modified default servers proposal. According to fate#323454 and bsc#1188980 we should propose a suse's pool and do not propose dhcp provided servers (but offer them for selection). So, the proposal looks like this:
ntp_patched_4

src/modules/NtpClient.rb
mchf: To summarize what I got from release manager(s)
- in general fine with the change
- wants to keep pool.conf in chrony-pool-* package because of image based installations
- fine with overwriting pool.conf from chrony-pool-* in case of yast driven installation
- having server directive in pool.conf was not problem

So, basically the patch should, i believe, fulfill the above.

src/modules/NtpClient.rb
```diff
@@ -384,6 +384,14 @@ def GetUsedNtpServers
       ntp_conf.pools.keys
     end
 
+    # @return [Hash<String, Symbol> pair of source address and type (server, pool)
+    def GetUsedNtpSources
+      sources = ntp_conf.servers.keys.each_with_object({}) { |s, res| res[s] = :server }
```
teclator, line 389:
```suggestion
     ntp_conf.servers.keys.each_with_object(sources = {}) { |s, res| res[s] = :server }
```

src/modules/NtpClient.rb
```diff
@@ -384,6 +384,14 @@ def GetUsedNtpServers
...
+      sources = ntp_conf.pools.keys.each_with_object(sources) { |s, res| res[s] = :pool }
```
teclator, line 390:
```suggestion
      ntp_conf.pools.keys.each_with_object(sources) { |s, res| res[s] = :pool }
```

teclator:
I tested it in a **Tumbleweed** installation with a **DUD** and from behavior **POV** it works as expected. It would be nice to cleanup it a little bit more and to cover it better with unit tests but we can leave it for next iteration.

teclator:
> > Just a minor suggestion, slightly out of the scope of this PR: why not take the opportunity to put _Current time_ and _Current date_ inline?
> > I mean, when selecting _Manually_ see something like
> > ```
> > Date        Time
> > ▒▒▒▒▒▒▒▒▒   ▒▒▒▒▒▒▒▒▒
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > instead of the current
> > ```
> >    
> > Current Time
> > ▒▒▒▒▒▒▒▒▒  
> > 
> > Current Date     
> > ▒▒▒▒▒▒▒▒▒
> > ```
> 
> it is in different module (yast2-country) see the description how it works to get into fun with this dialog ;-)

And doesn't make sense to change it in **yast2-country**? We will have more space for the sources table

teclator:
> I've modified default servers proposal. According to fate#323454 and bsc#1188980 we should propose a suse's pool and do not propose dhcp provided servers (but offer them for selection). So, the proposal looks like this: ntp_patched_4

np: It is a little bit strange that the table do not take the whole width.

src/clients/ntp-client_proposal.rb
```diff
@@ -124,7 +143,14 @@ def ui_enable_disable_widgets(enabled)
         else
           UI.ChangeWidget(Id(:ntp_now), :Enabled, enabled)
         end
+
         UI.ChangeWidget(Id(:ntp_save), :Enabled, enabled)
+        UI.ChangeWidget(Id(:ntp_address), :Enabled, enabled)
+
+        @sources_table.send(enabled ? :enable : :disable)
```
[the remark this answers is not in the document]
mchf: this piece of code is used from ```yast2-country``` to drive availability of the ntp piece of dialog. Basically, if radio button ```Synchronize with NTP Server``` is not checked, everything in the section should be disabled. That is driven by method which receives ``` enabled = true|false```.  ```send``` is used because it looked more pretty to me then ```if enabled then enable else disable end``` (or similar) approach.

And why ```enable|disable``` instead of ```enabled?``` ? Well AFAIK first changes the state of widged, the second just reports it.

mchf:
> > > Just a minor suggestion, slightly out of the scope of this PR: why not take the opportunity to put _Current time_ and _Current date_ inline?
> > > I mean, when selecting _Manually_ see something like
> > > ```
> > > Date        Time
> > > ▒▒▒▒▒▒▒▒▒   ▒▒▒▒▒▒▒▒▒
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > instead of the current
> > > ```
> > >    
> > > Current Time
> > > ▒▒▒▒▒▒▒▒▒  
> > > 
> > > Current Date     
> > > ▒▒▒▒▒▒▒▒▒
> > > ```
> > 
> > 
> > it is in different module (yast2-country) see the description how it works to get into fun with this dialog ;-)
> 
> And doesn't make sense to change it in **yast2-country**? We will have more space for the sources table

Sure it does. But cannot be done here.

src/clients/ntp-client_proposal.rb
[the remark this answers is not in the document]
teclator: > this piece of code is used from `yast2-country` to drive availability of the ntp piece of dialog. Basically, if radio button `Synchronize with NTP Server` is not checked, everything in the section should be disabled. That is driven by method which receives ` enabled = true|false`. `send` is used because it looked more pretty to me then `if enabled then enable else disable end` (or similar) approach.

Understood, but then it is probably to use **public_send** instead..
> 
> And why `enable|disable` instead of `enabled?` ? Well AFAIK first changes the state of widged, the second just reports it.

true

teclator (changes requested):
Please update the description of the PR with current state, fix the failing unit test and I guess you could also remove WIP from the title apart of add the changelog and bump version entries.

teclator (approved):
LGTM

PR #174, opened 2022-09-21
Skip netconfig configuration when it is not available (bsc#1198066)

teclator:
## Problem

During installation, we try to call **netconfig** for writing and updating the ntp policy even when **netconfig** it is not available.

- [bsc#1198066](https://bugzilla.suse.com/show_bug.cgi?id=1198066)

## Solution

Skip netconfig configuration if it is not available

## Testing

- *Added a new unit test*

### DIFF
--- a/test/ntp_client_test.rb
+++ b/test/ntp_client_test.rb
@@ -222,6 +222,8 @@ describe Yast::NtpClient do
   end
 
   describe "#Write" do
+    let(:netconfig) { true }
+
     before do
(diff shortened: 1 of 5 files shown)


test/ntp_client_test.rb
```diff
@@ -4,7 +4,6 @@
 require "cfa/memory_file"
 
 Yast.import "NtpClient"
```
teclator, line 6: It is not used anymore, so removed it.

PR #175, opened 2022-11-21
Fix netconfit path

teclator:
## Problem

In #174 we fixed a Tumbleweed bug avoiding to call **netconfig** when it was not present in the system but as consequence of that bug the executable is probably not called anymore in **SLE-15-SP5** as we modified the full path of the executable to be **/usr/sbin/netconfig** which is wrong as it should be **/sbin/netconfig**.

- [*bsc#1205401*](https://bugzilla.suse.com/show_bug.cgi?id=1205401)

## Solution

Fix the **netconfig full path** in **SLE-15-SP5** by now although we should look for a better solution in order to not use full paths anymore avoiding these kinds of problems **(usr-merge)**.


shundhammer (approved):
LGTM

